PR #41, opened 2024-04-04
feat: devnet deployment QOL improvements

MerkleBoy:
The framework is overall more resilient to changes between mono-namespace and multi-namespace deployment; some scripts have been added to help builders do basic things (more like that will come, that would be the focus of another PR thou)


0xxlegolas (commented):
Why there is a namespace constant `FRONTIER_WORLD_DEPLOYMENT_NAMESPACE` when there is a configurable namespace parameter ? 

In all the implementation having a namespace constant in the code does bother.
The namespace parameter should be able to handle any string which is passed either its a different namespace for different module or the same. Regardless of the string value the namespace implementation should not vary. 

```
function _namespace(bytes14 namespace) internal view returns (bytes14) {
    if (!ResourceIds.getExists(WorldResourceIdLib.encodeNamespace(namespace))) {
      return FRONTIER_WORLD_DEPLOYMENT_NAMESPACE;
    }
    return namespace;
  }
```

0xxlegolas (approved):

PR #52, opened 2024-04-23
Move global constants into their co-located constant files.

tinkrtailor:
This commit also moves the smart object framework constants from the common module into their respective local module for stronger modularity.


MerkleBoy (commented):
overall looks good !

0xxlegolas (approved):

PR #63, opened 2024-05-06
feat: SOF integration

MerkleBoy:
by the end of this feature branch we should have constrained the entire framework to leverage features in the Smart-Object framework. Non exhaustive list:

- ready-to-use `CLASS`es to tag `OBJECT`s with 
- gating functions to `entityId`s that are tagged with the right class
- easy detection of entity-to-class (and vice versa?)


MerkleBoy:
there is a bug with the deployment script that I can't pinpoint yet, something to do with symbolic links, imports, and setForwarder for some reason. It's present on develop also

mud-contracts/frontier-world/src/modules/smart-storage-unit/systems/SmartStorageUnit.sol
```diff
@@ -48,7 +51,7 @@ contract SmartStorageUnit is EveSystem {
    */
   function createAndAnchorSmartStorageUnit(
     uint256 smartObjectId,
-    EntityRecordData memory entityRecordData,
+    EntityRecordTableData memory entityRecordData,
```
0xxlegolas, line 54: this should be EntityRecordData

MerkleBoy: Oh I thought we needed to use the autogenerated schema for `EntityRecordData` ; there are other places where I made similar changes

Aren't they the same ?

mud-contracts/common/constants/src/constants.sol
```diff
@@ -15,6 +15,14 @@ bytes14 constant INVENTORY_DEPLOYMENT_NAMESPACE = "eveworld";
 bytes14 constant SMART_DEPLOYABLE_DEPLOYMENT_NAMESPACE = "eveworld";
 bytes14 constant SMART_STORAGE_UNIT_DEPLOYMENT_NAMESPACE = "eveworld";
 
+// placeholder values
+uint256 constant SMART_CHARACTER_CLASS_ID = 123456;
```
0xxlegolas, line 19: can we move to the env variables where it takes the Deployable, Character and Inventory typeId's and item Ids? We could get read the values form the .env and hash it to create the classIds by hashing 
eg: `uint256 classId1 = uint256(keccak256(abi.encodePacked("typeId12")));`

0xxlegolas:
on a high level everything look good to me. Will do some testing and approve.

mud-contracts/frontier-world/src/modules/smart-storage-unit/systems/SmartStorageUnit.sol
0xxlegolas: No, 
The autogenerated schemas are only for contract references. To keep the interfaces clean We define the struct  based on what inputs we need from the client and use that struct interface for external communication. 

Reason being 
1. we dont want to convolute the external interface struct with the internal mud table struct data 
eg: In mud.config 
```
 EntityRecordTable: {
      keySchema: {
        entityId: "uint256",
      },
      valueSchema: {
        itemId: "uint256",
        typeId: "uint256",
        volume: "uint256",
        recordExists: "bool",
      },
      tableIdArgument: true,
    },
```
The table definition is for on chain data struct. Here the `recordExists` value is mainly for internal tracking, The clients does not have to worry about passing this data according to the original design. There are many such examples where the table struct has more info which is not necessary for the clients. 

So instead of using the auto generated structs `EntityRecordTableData` its always clean to use a external interface struct with only the inputs we need from clients.

currently in many places we dont follow this but its huge refactoring. so I am just requesting changes only in places where its required. post play test we should make sure everything is consistent

MerkleBoy: fixed in aef1c73

mud-contracts/common/constants/src/constants.sol
MerkleBoy: I have added helper library functions in the ModulesRegistrationLibrary so that we can create SSU or Deployables or Smart Characters by directly using a `typeId` in input and it returns the hashed `classId` directly

It was a bit messy to put this there, agreed

0xxlegolas (approved):

PR #77, opened 2024-05-16
Rename org scope and fix version for first publish

gunnigylfa:


0xxlegolas (commented):
codegen folder can be added to `.gitignore`

0xxlegolas (approved):

gunnigylfa:
@0xxlegolas 
It seems we need the codegen folder because consumption of the contracts depends a lot on the generated code and interfaces: https://github.com/projectawakening/world-chain-contracts/blob/develop/end-to-end-tests/script/CreateAndAnchor.s.sol#L7 and https://github.com/projectawakening/world-chain-contracts/blob/develop/end-to-end-tests/script/CreateSmartCharacter.s.sol#L10



I also noticed that MUD includes this https://github.com/latticexyz/mud/tree/main/packages/world/src/codegen

PR #79, opened 2024-05-16
Ccp merkebloy/gatekeeper module

MerkleBoy:


mud-contracts/world/src/codegen/world/ISmartStorageUnit.sol
```diff
@@ -16,7 +17,7 @@ interface ISmartStorageUnit {
 
   function eveworld__createAndAnchorSmartStorageUnit(
     uint256 smartObjectId,
-    EntityRecordData memory entityRecordData,
+    EntityRecordTableData memory entityRecordData,
```
0xxlegolas, line 20: I dont think this should be changed. external Interface requires `EntityRecordData`

MerkleBoy: that was an artifact, sorry
changing this now

PR #81, opened 2024-05-17
inventory bug fix

0xxlegolas:
- added a new parameter ephemeralOwner to transferFromInventoryEphemeral function so that it know to which inventory to transfer to. 
- fixed quantity check inclusive condition


MerkleBoy (commented):
is that a temporary bugfix before we use `initialMsgSender()` as the default address to use for ephemeral inventory functions ?

0xxlegolas:
> is that a temporary bugfix before we use `initialMsgSender()` as the default address to use for ephemeral inventory functions ?

Yes,

mud-contracts/frontier-world/src/codegen/world/IInventoryInteract.sol
ccp-tezza, line 1: we shouldn't be tracking codegen files, right?

fc-tezza (approved):
Updates look good

mud-contracts/frontier-world/src/codegen/world/IInventoryInteract.sol
MerkleBoy: oh it might be something we need to track now, @gunnigylfa added auto-publishing of ABIs on the blockchain gateway that might be related . not 100% sure though

0xxlegolas: Yes we need for publishing

MerkleBoy (approved):
amazing 👏 
LGTM !

PR #83, opened 2024-05-18
Access Control Module

fc-tezza:
Features -
AccessControl (World coupled): 

- create singleton account ROOT roles that can act as ADMIN for themselves and other created roles. Bootstrapping and decentralising ADMIN role creation and moving away from the centralised Ownable pattern
- manage memberships of created roles by granting/revoking/renouncing role membership to any account from an ADMIN permissioned account
- check account membership for any created role 
- check if a role exists and if it is a ROOT role or not

AccessRulesConfig (World coupled):
- set a list of roles to be checked against any on-chain activity for two account contexts: ORIGIN (tx.origin) and TRANSIENT (world().intialMsgSender). ORIGIN tracks the signer and submitter of a transaction, while TRANSIENT tracks the assigned actor of a transaction (in our case the ERC2771 MetaTxn signer and primary interacting account for users).
- set a level of enforcement for access checking, ORIGIN, TRANSIENT, or both
- multiple configurations per Entity are available for maximum flexibility

AccessRuleMock (decoupled from the World deployment):
- an example hook logic implementation that can be attached to any function in the Smart Object Framework which implements the `hookable()` modifier. Protect any function for your created Entities.
- point at a System/function pair, assign which entity to enforce your Role Based Access rules for, assign an enforcement configuration and lock it up!


MerkleBoy (approved):
LGTM; the part about method visibility is just a suggestion, in a way we can read those values directly from the indexer so it's not that big of a deal

ccp-tezza:
Closing this since it is stale , will need rework for SOF v2, and we have another stop-gap access control PR coming down the pipeline.

ccp-tezza:
Closing this since it is stale , will need rework for the SOF v2, and we have another stop-gap access control PR coming down the pipeline.

PR #86, opened 2024-05-20
adding a stable release

0xxlegolas:


gunnigylfa:
Two things:
First release branched for versioned releases should not be reused. We should have them versioned.
Second why are there so many changes here? A stable release branch should not be divergent in logic it should be a point in time copy of our code and be ephemeral.

end-to-end-tests/package.json
```diff
@@ -45,9 +49,9 @@
     "@latticexyz/world": "2.0.0-transaction-context-98ef570f",
     "@latticexyz/world-modules": "2.0.0-transaction-context-98ef570f",
     "@latticexyz/gas-report": "2.0.0-transaction-context-98ef570f",
-    "@eveworld/smart-object-framework": "0.0.1",
-    "@eveworld/world": "0.0.1",
-    "@eveworld/common-constants": "0.0.1"
+    "@eveworld/smart-object-framework": "0.0.2",
```
gunnigylfa, line 52: Lets remove these changes since develop now has a direct link to the source packages

gunnigylfa:
I guess this is now just  the withdraw bug fix?

PR #110, opened 2024-06-05
inventory interact inventoryToEphemeralTransferWithParam with paramater

0xxlegolas:
bug fix - inventory interact inventoryToEphemeralTransferWithParam with paramater

### DIFF
--- a/mud-contracts/world/src/modules/inventory/systems/InventoryInteract.sol
+++ b/mud-contracts/world/src/modules/inventory/systems/InventoryInteract.sol
@@ -85,8 +85,8 @@ contract InventoryInteract is EveSystem {
   }
 
   /**
-   * @notice Transfer items from ephemeral to inventory
-   * @dev transfer items from ephemeral to inventory
+   * @notice Transfer items from inventory to ephemeral
+   * @dev transfer items from inventory to ephemeral
    * @param smartObjectId is the smart object id
    * @param ephemeralInventoryOwner is the ephemeral inventory owner
    * @param outItems is the array of items to transfer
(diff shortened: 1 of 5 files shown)


mud-contracts/world/src/modules/inventory/systems/InventoryInteract.sol
```diff
@@ -84,6 +84,64 @@ contract InventoryInteract is EveSystem {
     _inventoryLib().depositToEphemeralInventory(smartObjectId, ephemeralInventoryOwner, inItems);
   }
 
+  /**
+   * @notice Transfer items from ephemeral to inventory
+   * @dev transfer items from ephemeral to inventory
```
fc-tezza: These comments are incorrect, this is a transfer from inventory to ephemeral

mud-contracts/world/src/modules/inventory/systems/InventoryInteract.sol
```diff
@@ -84,6 +84,64 @@ contract InventoryInteract is EveSystem {
...
+   * @param smartObjectId is the smart object id
+   * @param ephemeralInventoryOwner is the ephemeral inventory owner
+   * @param outItems is the array of items to transfer
+   */
+  function inventoryToEphemeralTransferWithParam(
+    uint256 smartObjectId,
+    address ephemeralInventoryOwner,
```
ccp-tezza, line 96: For interfacing this is clearer, but technically we can just use the parameter in the InventoryItem object (as now it is not being used). In any case, I am fine to leave it as is for now.

fc-tezza (approved):
Just a few sideline comments, but overall looks good to me.

mud-contracts/world/src/modules/inventory/systems/InventoryInteract.sol
0xxlegolas: fixed

PR #112, opened 2024-06-13
bugfix: fix `_removeEntityHookAssociation()`

MerkleBoy:
sneaky one


fc-tezza (commented):
All looks good to me.. I suggest that we merge @gunnigylfa 's PR fist https://github.com/projectawakening/world-chain-contracts/pull/113

And then reflect those changes here with a merge from develop so this also passes all CI/CD checks

PR #116, opened 2024-07-03
EF-4558: Set immediate exit flags on errors in bash script steps

tinkrtailor:
Also turns onerror flags that error out when the script:
-  References variables that are not set
-  The script is used in a pipeline and it will make the whole pipe fail.


ccp-error (approved):
LGTM!

0xxlegolas (approved):

PR #120, opened 2024-07-12
fix: Only build amd architecture images for the deployer image

tinkrtailor:
Foundry images do not support building for ARM architectured platforms. Until that is fixed we will only provide a deployer image with that platform.


0xxlegolas (approved):

0xxlegolas:
@gunnigylfa  can we merge this?

tinkrtailor:
@0xxlegolas yes we can. Merge one checks sucessfully pass

PR #122, opened 2024-07-22
refactor: world v2 entity record

0xxlegolas:
**Description**

- simplified implementation for world-v2 
- entity record functions


mud-contracts/world-v2/src/systems/constants.sol
```diff
@@ -0,0 +1,4 @@
+//SPDX-License-Identifier: MIT
```
tinkrtailor, line 1: Should this file be at the top level of the source code directory?

mud-contracts/world-v2/src/systems/entity-record/EntityRecordSystem.sol
```diff
@@ -0,0 +1,65 @@
...
+   * @dev sets the name of an entity
+   * @param entityId the id of a in game entity referred as smart object id
+   * @param name the name of the entity
+   */
+  function setName(uint256 entityId, string memory name) public {
+    EntityRecordMetadata.setName(entityId, name);
+  }
+
+  /**
+   * @dev sets the description of an entity
```
tinkrtailor: Is this doc comment correct?

mud-contracts/world-v2/mud.config.ts
```diff
@@ -1,13 +1,34 @@
...
+        recordExists: "bool",
       },
-      key: [],
+      key: ["entityId"],
+    },
+    EntityRecordMetadata: {
+      schema: {
+        entityId: "uint256",
+        name: "string",
+        dappURL: "string",
```
tinkrtailor, line 28: Why does every entity have a dappurl in its metadata?

0xxlegolas: Each entity can have an associated external DApp. For example, any entity that can be smart-deployed or act as a smart turret, when customized by the builder, requires a URL to access the custom-built DApp.

tinkrtailor (approved):

PR #138, opened 2024-08-28
feat: kill mail

ccp-error:
Support for storing kill mails on chain.

Resolves https://ccpgames.atlassian.net/browse/EF-5113.


mud-contracts/world/mud.config.ts
```diff
@@ -392,5 +393,22 @@ export default mudConfig({
...
+
+    KillMailTable: {
+      keySchema: {
+        killMailId: "uint256",
+      },
+      valueSchema: {
+        killer: "address",
+        victim: "address",
+        loss: "KillMailLoss",
+        timestamp: "uint256",
```
0xxlegolas: can we rename to make it more intuitive. What timestamp is this ?

mud-contracts/world/src/modules/kill-mail/constants.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+pragma solidity >=0.8.21;
+
+bytes16 constant KILL_MAIL_SYSTEM_NAME = "KillMail";
+
+bytes16 constant KILL_MAIL_MODULE_NAME = "KillMail";
+bytes14 constant KILL_MAIL_MODULE_NAMESPACE = "KillMail";
+
+bytes16 constant KILL_MAIL_TABLE_NAME = "KillMail";
```
0xxlegolas:
```suggestion
bytes16 constant KILL_MAIL_TABLE_NAME = "KillMailTable";
```

mud-contracts/world/src/modules/kill-mail/systems/KillMail.sol
```diff
@@ -0,0 +1,38 @@
...
+import { IKillMailErrors } from "../IKillMailErrors.sol";
+import { Utils } from "../Utils.sol";
+
+contract KillMail is EveSystem {
+  function reportKill(
+    uint256 killMailId,
+    address killer,
+    address victim,
+    KillMailLoss loss,
+    uint256 timestamp
```
0xxlegolas: is this is the timestamp the kill was reported or the time its stored on-chain ?

0xxlegolas (commented):
Looks good so far ! Great

mud-contracts/world/mud.config.ts
ccp-error: Yes definitely.

It's the timestamp of when the kill happened in game.

mud-contracts/world/src/modules/kill-mail/constants.sol
ccp-error: Good catch!

mud-contracts/world/src/modules/kill-mail/systems/KillMail.sol
ccp-error: The in-game timestamp for the kill.

mud-contracts/world/mud.config.ts
ccp-error: Thinking about changing it to `killTimestamp` or `killTime` or something like that.

What do you think about that @0xxlegolas ?

mud-contracts/world/src/modules/kill-mail/systems/KillMail.sol
ccp-error: Changed to `killTimestamp`.

mud-contracts/world/mud.config.ts
ccp-error: Changed to `killTimestamp`.

mud-contracts/world/src/modules/kill-mail/systems/KillMail.sol
```diff
@@ -0,0 +1,38 @@
...
+  function reportKill(
+    uint256 killMailId,
+    address killer,
+    address victim,
+    KillMailLossType lossType,
+    uint256 killTimestamp
+  ) public hookable(killMailId, _systemId()) {
+    ResourceId tableId = Utils.killMailTableId(_namespace());
+
+    if (KillMailTable.getKiller(tableId, killMailId) != address(0)) {
```
vayan, line 19: we're getting the whole kill 
```suggestion
    if (KillMailTable.getKill(tableId, killMailId) != address(0)) {
```

ccp-error: This is just to check if a kill has been registered or not.

vayan: Oh I was confused, so shouldn't it be:
```suggestion
    if (KillMailTable.getKiller(tableId, killMailId) == address(0)) {
```

Or maybe I'm missing something

vayan: oh `address(0)` is not the `killer` argument 🤦 nevermind, all good

mud-contracts/world/src/codegen/common.sol
```diff
@@ -15,3 +15,8 @@ enum SmartAssemblyType {
   SMART_TURRET,
   SMART_GATE
 }
+
+enum KillMailLossType {
+  SHIP,
+  POD
+}
```
vayan, line 22: can you override the enum values in solidity? else we should maybe add the "UNSPECIFIED" type to match the proto enum.

```suggestion
enum KillMailLossType {
  UNSPECIFIED,
  SHIP,
  POD
}
```

the enum in protobuf: 

```proto
enum LossType {
  LOSS_UNSPECIFIED = 0;
  SHIP = 1;
  POD = 2;
}
```

ccp-error: Yes I did think about that.

However, the unspecified part of the proto is there because proto "requires" it. (https://protobuf.dev/programming-guides/dos-donts/#unspecified-enum)

We should never store kill mails that have an unspecified loss type.

vayan: yeah and we never will, but it was mostly to have the same enum value everywhere. 
Else we'll have `SHIP=0` in indexer/chain and `SHIP=1` in game/protobuf.

if we can do stuff like that it also works: 
```suggestion
enum KillMailLossType {
  SHIP=1,
  POD=2
}
```

Just a nitpick anyway 👍

ccp-error: I understand the reasoning.

I think we should do a mapping in the code the reads from the contract. The reason for that is that if the enum order is changed in one place but not the other (someone might forget to do it in both places, might not know about the proto definitions, etc.), then it will break the system.

mud-contracts/world/src/modules/kill-mail/systems/KillMail.sol
```diff
@@ -0,0 +1,29 @@
...
+      revert IKillMailErrors.KillMail_AlreadyExists("Can't report an already reported kill");
+    }
+
+    KillMailTable.set(tableId, killMailId, killMailTableData);
+  }
+
+  function _systemId() internal view returns (ResourceId) {
+    return Utils.killMailSystemId(_namespace());
+  }
+}
```
0xxlegolas, line 29: @ccp-error  Do we need a function to delete or update the killMail. Or once its reported it cant be changed ?

0xxlegolas (commented):
LGTM

mud-contracts/world/src/modules/kill-mail/systems/KillMail.sol
ccp-error: I do not think we need a function to update or delete the kill mail.

0xxlegolas (approved):

ccp-raudur (commented):
Wow

vayan (approved):
good job @ccp-error proud of you.

PR #140, opened 2024-09-03
refactor: smart turret interfaces

0xxlegolas:
- updated function interfaces to match protobuf definition : https://github.com/ccpgames/eve-frontier-proto/blob/main/eve/blockchain/smartturret/smartturret.proto
- added function to updateCorpId


vayan (approved):
cleaner 🙏

fc-tezza (approved):
This interface is much cleaner! All changes look good!

PR #142, opened 2024-09-04
chore: update CODEOWNERS with team

vayan:
use team instead of individual accounts


vayan:
can we give @projectawakening/codeowners explicit write access to the repository? then it should work :)

tinkrtailor (approved):

PR #144, opened 2024-09-05
improvement:"Add parent flag for mkdir within the deployer image

tinkrtailor:


ccp-bofai (approved):
Thank you! That should work for when any of these are volume mounted now :)

0xxlegolas (approved):

PR #145, opened 2024-09-09
chore: add scripts for creating smart gate

rhubarbgarden:
add script to run `createAndAnchorGate:devnet` from cli


0xxlegolas:
@rhubarbgarden  There is already script to do the same `createAndAnchorGate` : https://github.com/projectawakening/world-chain-contracts/blob/develop/end-to-end-tests/script/ConfigureSmartGate.s.sol#L76. Is there a reason to include this in end-to-end tests again ?

rhubarbgarden:
It's sometimes useful for front-end testing to be able to create smart gates on their own without going through the configure/link process.

0xxlegolas:
These scripts are mainly for end-to-end testing to ensure the contracts function as expected. Instead of duplicating them for front-end testing, we can refactoring them into smaller parts to avoid the configure/link process and make them more flexible. 
Will have the correction in the next PR ! Ref: https://ccpgames.atlassian.net/browse/EF-5201

PR #147, opened 2024-09-11
Chore: Create dependabot.yml

tinkrtailor:
Dependabot has a hard time understanding monorepos with multiple package.json files. Which is why we are configuring a checker for each of our packages.


0xxlegolas:
does it check for mud npm package updates as well ?

tinkrtailor:
@0xxlegolas I know that it direct dependencies. I think it also checks transitive dependencies via the lock file. But it only opens PRs for our own dependencies.

0xxlegolas:
Thanks for the clarification!

0xxlegolas (approved):

PR #175, opened 2024-09-13
smart gate feature implementation

0xxlegolas:
- Checking the maxDistance before linking 2 gates

### DIFF
--- a/mud-contracts/world/src/codegen/world/ISmartGate.sol
+++ b/mud-contracts/world/src/codegen/world/ISmartGate.sol
@@ -16,7 +16,7 @@ interface ISmartGate {
   error SmartGate_NotConfigured(uint256 smartGateId);
   error SmartGate_GateAlreadyLinked(uint256 sourceGateId, uint256 destinationGateId);
   error SmartGate_GateNotLinked(uint256 sourceGateId, uint256 destinationGateId);
-  error SmartGate_GateNotWithtinRange(uint256 sourceGateId, uint256 destinationGateId);
+  error SmartGate_NotWithtinRange(uint256 sourceGateId, uint256 destinationGateId);
 
   function eveworld__createAndAnchorSmartGate(
     uint256 smartGateId,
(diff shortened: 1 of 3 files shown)


mud-contracts/world/src/modules/smart-gate/systems/SmartGate.sol
```diff
@@ -123,6 +126,10 @@ contract SmartGate is EveSystem, AccessModified {
 
     //TODO: Check if the state is online for both the gates ??
     //TODO: Link the gates only when the distance between 2 gates are less than the max distance
```
vayan:
```suggestion
```
Done now 😄

mud-contracts/world/src/codegen/world/ISmartGate.sol
```diff
@@ -42,5 +43,5 @@ interface ISmartGate {
 
   function eveworld__isGateLinked(uint256 sourceGateId, uint256 destinationGateId) external view returns (bool);
 
-  function eveworld__calculatDistance(uint256 sourceGateId, uint256 destinationGateId) external view returns (uint256);
+  function eveworld__isWithinRange(uint256 sourceGateId, uint256 destinationGateId) external view returns (bool);
```
vayan, line 46:
```suggestion
  function eveworld__isGateWithinRange(uint256 sourceGateId, uint256 destinationGateId) external view returns (bool);
```

after using `eveworld__inProximity` and `eveworld__agression` I think we should be a bit more verbose

0xxlegolas: Yes good catch, Will recompile it and push. Thanks

fc-tezza (approved):
All looks good to me..

My only comment is that to be very safe in the contract logic, in the future we may wan to ensure that the maximum possible distance^2 doesn't not overflow uint256 (and we may want to enforce this on the parameter inputs on the contract level). But other than this, all looking great for the current release.

PR #179, opened 2024-09-17
Refactor Gate Table Structure

0xxlegolas:
- added condition to check source and destination is not same 
- refactored gate table to make a single key instead of 2 key for query
- added 2 way connection for source and destination


mud-contracts/world/src/modules/smart-gate/systems/SmartGate.sol
```diff
@@ -124,12 +125,18 @@ contract SmartGate is EveSystem, AccessModified {
...
+    }
+
     //TODO: Check if the state is online for both the gates ??
     if (isWithinRange(sourceGateId, destinationGateId) == false) {
       revert SmartGate_NotWithtinRange(sourceGateId, destinationGateId);
     }
 
+    //Create a 2 way link between the gates
     SmartGateLinkTable.set(_namespace().smartGateLinkTableId(), sourceGateId, destinationGateId, true);
+    SmartGateLinkTable.set(_namespace().smartGateLinkTableId(), destinationGateId, sourceGateId, true);
```
vayan, line 139: 🧡 thanks

vayan (approved):

ccp-raudur (approved):

PR #182, opened 2024-09-18
v0.0.14/naming convention

0xxlegolas:
- renaming `smartStorageUnitId` to  `smartObjectId` for inventory for consistency


tinkrtailor (approved):

ccp-error:
Are we changing the name back?

Why not change it in other places to the new name instead?

0xxlegolas:
@gunnigylfa  proper naming conventions across all the systems are implemented in v2. However in this version if we change the name its a huge change from the game play. So to minimize the change from game we are renaming it now

ccp-error:
Would have appreciated a reply to my comment before this being merged.

PR #183, opened 2024-09-18
v0.0.14/naming convention

0xxlegolas:
- renaming for `smartObjectId` across systems and MUD Table for consistency

### DIFF
--- a/mud-contracts/world/src/modules/smart-gate/systems/SmartGate.sol
+++ b/mud-contracts/world/src/modules/smart-gate/systems/SmartGate.sol
@@ -47,7 +47,7 @@ contract SmartGate is EveSystem, AccessModified {
   using Utils for bytes14;
 
   error SmartGate_UndefinedClassId();
-  error SmartGate_NotConfigured(uint256 smartGateId);
+  error SmartGate_NotConfigured(uint256 smartObjectId);
   error SmartGate_GateAlreadyLinked(uint256 sourceGateId, uint256 destinationGateId);
   error SmartGate_GateNotLinked(uint256 sourceGateId, uint256 destinationGateId);
   error SmartGate_NotWithtinRange(uint256 sourceGateId, uint256 destinationGateId);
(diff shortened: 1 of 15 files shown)


ccp-raudur (approved):

mud-contracts/world/src/modules/smart-gate/systems/SmartGate.sol
```diff
@@ -153,22 +153,22 @@ contract SmartGate is EveSystem, AccessModified {
...
+  ) public onlyAdminOrObjectOwner(smartObjectId) hookable(smartObjectId, _systemId()) {
+    SmartGateConfigTable.setSystemId(_namespace().smartGateConfigTableId(), smartObjectId, systemId);
   }
 
   /**
    * @notice view function for smart gates which is linked
-   * @param characterId is of the id of the character
-   * @param sourceGateId is the id of the source gate
-   * @param destinationGateId is the id of the destination gate
+   * @param characterId is of the smartObjectId of the character
```
ccp-error, line 169: Should we treat all ids as smart object ids? So a character id will not be simply the character's integer as uint256?

ccp-error (approved):
Love it! Thank you!

Should we treat all ids as smart object ids, even ids like character ids?

So far I think we've been treating character ids as just an uint256 representation of the in-game character id.

mud-contracts/world/src/modules/smart-gate/systems/SmartGate.sol
0xxlegolas: I am afraid this might be a huge change for the game play. May be next iteration ?

ccp-error: Yeah I think so. I think we should for now just change everything that is being generated as smart object id **and** are primary keys of said objects, to be called `smartObjectId`.

0xxlegolas:
> Love it! Thank you!
> 
> Should we treat all ids as smart object ids, even ids like character ids?
> 
> So far I think we've been treating character ids as just an uint256 representation of the in-game character id.

Ideally Yes. But not sure for this release as this might affect game client and we want to keep the changes less for them

mud-contracts/world/src/modules/smart-gate/systems/SmartGate.sol
0xxlegolas: We cant do this in some cases eg: 
```
function linkSmartGates(uint256 sourceGateId, uint256 destinationGateId)
```
Thought both are the smartObjectIds we cannot have the same naming. So in some places its okay to have descriptive names I suppose. WDYT ?

ccp-error:
> > Love it! Thank you!
> > Should we treat all ids as smart object ids, even ids like character ids?
> > So far I think we've been treating character ids as just an uint256 representation of the in-game character id.
> 
> Ideally Yes. But not sure for this release as this might affect game client and we want to keep the changes less for them

Ah ok I see. Yes I don't think we should change the data of these fields in this release.

mud-contracts/world/src/modules/smart-gate/systems/SmartGate.sol
ccp-error: Yes exactly. In this case these are not primary keys, more like foreign keys.

I guess for "primary keys" we should strive to use `smartObjectId` (usually for primary keys I just use `id` but since we've been using `smartObjectId` thus far I think that's alright).

0xxlegolas: cool, so `characterId` to `smartObjectId`  change in this iteration or next ?

ccp-error: I think `characterId` can keep its name and still be just a normal integer (not of the type `smartObjectId`).

PR #184, opened 2024-09-19
fix: killer and victim values should be char ids rather than addr

ccp-error:
- killer changed from address to character id (smart object id)
- victim changed from address to character id (smart object id)

Argument names and table field names changed.

- killer -> killerCharacterId
- victim -> victimCharacterId

### DIFF
--- a/mud-contracts/world/src/modules/kill-mail/systems/KillMail.sol
+++ b/mud-contracts/world/src/modules/kill-mail/systems/KillMail.sol
@@ -16,7 +16,7 @@ contract KillMail is AccessModified, EveSystem {
   ) public onlyAdmin hookable(killMailId, _systemId()) {
     ResourceId tableId = Utils.killMailTableId(_namespace());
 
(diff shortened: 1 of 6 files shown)


mud-contracts/world/src/modules/kill-mail/systems/KillMail.sol
```diff
@@ -16,7 +16,7 @@ contract KillMail is AccessModified, EveSystem {
   ) public onlyAdmin hookable(killMailId, _systemId()) {
     ResourceId tableId = Utils.killMailTableId(_namespace());
 
-    if (KillMailTable.getKiller(tableId, killMailId) != address(0)) {
+    if (KillMailTable.getKillerCharacterId(tableId, killMailId) != 0) {
```
ccp-error, line 19: @0xxlegolas this is the right way to check if a kill is already registered, right?

0xxlegolas: Yes this should work

0xxlegolas (approved):
LGTM

PR #188, opened 2024-09-23
improvement: quality of life improvements for InventoryInteract builder flows

fc-tezza:
- Enforce one-to-one mapping for smart character address to id - includes a new error - `SmartCharacter_AlreadyCreated(address characterAddress, uint256 characterId);`
- Added a reverse lookup table called `CharctersByAddressTable` - key characterAddress, returns characterId - used to enforce that any Ephemeral Inventory Owner that get sent items is in-fact a created character (so items don't get sent to blackhole addresses)
- Created a new simpler struct `TransferItem`, to make InventoryInteract usability easier
- `TransferItem` only includes: the `inventoryItemId`, `owner`, and `quantity` since the other fieilds are already in `EntityRecord`
- Subsequently use EntityRecord data to create the full InventoryItem object to pass into the Inventory and EphemeralInventory modules
- Implemented a `InventoryInteract.setApprovedAccessList` - this allows the Inventory owner to APPROVE others access to the InventoryInteract transfer functions via `_msgSender()` checking (including other Systems or EOAs)
- Implemented `setAllTransferAccess`, `setEphemeralToInventoryTransferAccess`, and `setInventoryToEphemeralTransferAccess` - these allow the Inventory Owner (`onlyOwner)` to toggle the access control on/off for the InventoryInteract.transfer functions.
- Enforcement modifier `onlyOwnerOrSystemApproved` for `InventoryInteract.inventoryToEphemeralTransfer` 
- When`onlyOwnerOrSystemApproved` is active, the Inventory Owner can still call directly to the function or through other System calls ( because `_initialMsgSender `is checked), however, other EOAs can only call by being APPROVED and calling direct OR they can cal via an APPROVED System (`_msgSender` is checked). This gives the Inventory Owner freedom to call how they wish but protect the `InventoryInteract.inventoryToEphemeralTransfer` function from 3rd party calls (unless APPROVED or calling via an APPROVED System).
- Enforcement rmodifier `onlySystemApproved` for `InventoryInteract.ephemeralToInventoryTransfer`
-  When `onlySystemApproved` is active, the Ephemeral Inventory Owner can no longer call directly unless they are APPROVED or call via and APPROVED System. This give the Object Inventory Owner the ability to control how Ephemeral Owners interact with their Object.


mud-contracts/world/src/modules/inventory/systems/EphemeralInventorySystem.sol
```diff
@@ -238,6 +247,14 @@ contract EphemeralInventorySystem is AccessModified, EveSystem {
     InventoryItem memory item,
     uint256 usedCapacity
   ) internal returns (uint256) {
+    // if (ephemeralInventoryOwner != item.owner) {
```
0xxlegolas, line 250: any reason why this check is not needed ?

mud-contracts/world/src/modules/smart-gate/systems/SmartGateSystem.sol
```diff
@@ -150,7 +150,6 @@ contract SmartGateSystem is EveSystem, AccessModified {
       revert SmartGate_GateNotLinked(sourceGateId, destinationGateId);
     }
     SmartGateLinkTable.set(sourceGateId, destinationGateId, false);
-    SmartGateLinkTable.set(destinationGateId, sourceGateId, false);
```
0xxlegolas: why is this line removed ?

mud-contracts/world/src/modules/smart-gate/systems/SmartGateSystem.sol
```diff
@@ -228,7 +227,7 @@ contract SmartGateSystem is EveSystem, AccessModified {
 
     // Sum of squares (distance squared in meters)
     uint256 distanceSquaredMeters = (dx * dx) + (dy * dy) + (dz * dz);
-    return distanceSquaredMeters <= (maxDistance * maxDistance);
+    return distanceSquaredMeters <= maxDistance;
```
0xxlegolas:
```suggestion
    return distanceSquaredMeters <= (maxDistance * maxDistance);
```

0xxlegolas (commented):
Few comments, rest looks good to be merged

mud-contracts/world/src/modules/inventory/systems/EphemeralInventorySystem.sol
fc-tezza: I left them out to include only essential parameter requirement checking for this round.. we can turn these on in a future PR!

mud-contracts/world/src/modules/smart-gate/systems/SmartGateSystem.sol
fc-tezza: A branch merge mistake. Good catch!

mud-contracts/world/src/modules/smart-gate/systems/SmartGateSystem.sol
fc-tezza: Another merge overwrite artifact, I will fix, thanks!

mud-contracts/world/src/modules/smart-gate/systems/SmartGateSystem.sol
fc-tezza: fixed in commit d01a2efbd7b109237c7b0efae4f1b592c54dc60c

mud-contracts/world/src/modules/smart-gate/systems/SmartGateSystem.sol
fc-tezza: fixed in commit d01a2efbd7b109237c7b0efae4f1b592c54dc60c

0xxlegolas (approved):

PR #206, opened 2024-09-26
fixing aggression bug

0xxlegolas:
- fixing aggression bug by returning the same input queue if the aggressor belongs to the same corp as owner


ccp-raudur (approved):
🙏

PR #296, opened 2024-10-24
feat: Smart Object Framework - Execution Context

fc-tezza:
A PR to add Execution Context tracking to the Smart Object Framework:
 - Execution Context tracking within the World transient storage
 - for every call we now track the `systemId`, `functionSelector`,  `_msgSender()`,  and `_msgValue()` in transient storage for retrieval and usage later in the execution
 - these values can be accessed via the `world.getWorldCallContext()` getter functions
 - you can also check the current "call depth" using `world.getCallCount()` which will return the total number of World calls that have been invoked during the current transaction
 - in addition to execution context tracking, we have added a new `context()` modifier which should be used in all non-root/non-view SOF integrated functions. This modifier ensures that any call to a state changing function is in-fact executed through the World call APIs, thereby standardising context variable behavior and calling patterns within the SOF.
 - added quality of life internal functions for easy access to tracked context variables: `_callMsgSender()`, `_callMsgSender(uint callCount)`,  `_callMsgValue()`, `_callMsgValue(uint callCount)`, `_callDataWithoutContext(bytes memory callDataWithContext)`
 - fixed a bug that prevented direct `staticcall`s through the world interface calls (you can now call `view` functions uninhibited!)
 - restructured file paths and System naming for the latest version of MUD
 - upgraded to the latest version of MUD


mud-contracts/smart-object-framework-v2/src/WorldWithContext.sol
```diff
@@ -0,0 +1,589 @@
...
+  }
+
+  function _trackWorldCallContextBefore(address msgSender, ResourceId systemId, bytes4 functionSelector) internal {
+    bytes32 trackedSystemId = ResourceId.unwrap(systemId);
+    bytes32 trackedMsgSender = bytes32(uint256(uint160(msgSender)));
+    uint256 trackedMsgValue = msg.value;
+    assembly {
+      let callCount := tload(0)
+      let sysIdSlot := add(mul(callCount, 4), 1)
+      let functIdSlot := add(mul(callCount, 4), 2)
```
0xxlegolas, line 521: is it required to calculate the slot index  each time. Since we know its 4 slots can't we increment the slot index ?

mud-contracts/smart-object-framework-v2/src/inherit/SmartObjectFramework.sol
```diff
@@ -0,0 +1,155 @@
...
+    return msgSender;
+  }
+
+  /**
+   * @notice Gets the tracked msg.sender for a specific world call context
+   * @dev Retrieves msg.sender for the specified call count in the world execution context stack
+   * @param callCount The specific call count to get the call context for
+   * @return address The tracked msg.sender address for a world call
+   */
+  function _callMsgSender(uint256 callCount) internal view returns (address) {
```
0xxlegolas, line 98: this is going to be very helpful

mud-contracts/smart-object-framework-v2/src/namespaces/evefrontier/systems/entity-system/EntitySystem.sol
```diff
@@ -5,48 +5,52 @@ import { IWorldKernel } from "@latticexyz/world/src/IWorldKernel.sol";
...
       expected[0] = ENTITY_CLASS;
-      revert IErrors.WrongEntityType(classId.getType(), expected);
+      revert WrongEntityType(classId.getType(), expected);
     }
     if (Classes.getExists(classId)) {
-      revert IErrors.ClassAlreadyExists(classId);
+      revert ClassAlreadyExists(classId);
     }
 
     Classes.set(classId, true, new bytes32[](0), new bytes32[](0));
```
0xxlegolas, line 48: why is the systemTags set to `new bytes32[](0)` ? isn't it supposed to include the systemTags ?

0xxlegolas (approved):
Looks GREAT

mud-contracts/smart-object-framework-v2/src/namespaces/evefrontier/systems/entity-system/EntitySystem.sol
fc-tezza: The lines of code after this in the function set the `systemTags` accordingly via a `TagSystem.setSystemTags` call. 

This should be done because there are several tables that need to be updated when setting a new tag, not just the Classes table.

```
    IWorldKernel(_world()).call(
      TagSystemUtils.tagSystemId(),
      abi.encodeCall(ITagSystem.setSystemTags, (classId, systemTags))
    );
```

mud-contracts/smart-object-framework-v2/src/WorldWithContext.sol
fc-tezza: You could be right, it may be possible to accomplish this through incrementing. However, for a first cut implementation I implemented the current way to be absolutely certain that we are using the current call count accurately in our calculations. So essentially, erring on the side of guaranteed expected behavior. 

But we can revisit this later for sure for optimizing!

PR #340, opened 2024-12-09
V0.0.18

0xxlegolas:
- added ERC721 deployable and smart character token address to the artifacts as part of the deployer image output


standard-contracts/package.json
```diff
@@ -29,7 +29,7 @@
     "@latticexyz/store": "2.2.8",
     "@latticexyz/world": "2.2.8",
     "@eveworld/world": "0.0.17",
-    "@eveworld/world-core": "0.0.2",
```
tinkrtailor, line 32: is it possible not to link?

0xxlegolas: This package is not published in npm since its a internal package. 
if this is published yes we could use the released one

tinkrtailor: nx should resolve the local one

tinkrtailor (approved):
LGTM. The regex-es scare me quite a bit

standard-contracts/package.json
0xxlegolas: it was weird how the ci/cd worked in the previous build without the package being published

0xxlegolas:
agreed! from v2 it can be written directly to json file from the scripts instead of console.logs

PR #341, opened 2024-12-09
feat: Smart Object Framework V2 - Access Control

fc-tezza:
A comprehensive PR which adds an access control layer to the Smart Object Framework for standardized usage everywhere:
- updated the SOF with the `access` modifier and accompanying logic 
- added an `enforceCallCount` modifier for better security practice. This can be use to ensure that a call was made directly to the MUD World rather than internally from another system. When used in parallel with the `scope` modifier, these ensure that execution context value are only used by 'trusted' systems
- updated the `scope` modifier to now enforce that any previous call was also in a scoped System, thereby ensuring that the call chain cannot leave scope and then come back into scope (security update)
- added a new `world.callStatic` call to the WorldWithContext. This allows for guaranteed execution of `staticcall` functions between System-to-System calls (without any transient storage interruptions), and also allows us to enforce that access rule implementations MUST BE `staticcall` functions
- implemented a standalone `RoleManagementSystem` which can be used (for any use case) to create roles and manage the members of those roles
- implemented `AccessConfigSystem`, to allow System namespace owners the ability to configure access control per System/function based on Class (or Object) level access rule requirement and toggle enforcement of these access rules
- implemented `SOFAccessSystem` to create the access rules for `EntitySystem` and `TagSystem`
- implemented two scripts for configuring the `EntitySystem` and `TagSystem` access rules
- unit tests and documentation (in `mud-contracts/smart-object-framework-v2/README.md`)


0xxlegolas:
@ccp-tezza  Can we resolve the merge conflicts ?

ccp-tezza:
> @ccp-tezza Can we resolve the merge conflicts ?

Done!

0xxlegolas (approved):

PR #347, opened 2024-12-19
fix gate link/unlink issue

0xxlegolas:
Fix link/unlink issue by adding a condition to prevent linking if either the source or destination gate is already linked

### DIFF
--- a/mud-contracts/world/src/modules/smart-gate/systems/SmartGateSystem.sol
+++ b/mud-contracts/world/src/modules/smart-gate/systems/SmartGateSystem.sol
@@ -121,7 +121,7 @@ contract SmartGateSystem is EveSystem, AccessModified {
    * @param destinationGateId is the smartObjectId of the destination gate
    */
   function linkSmartGates(uint256 sourceGateId, uint256 destinationGateId) public onlyAdminOrObjectOwner(sourceGateId) {
-    if (isGateLinked(sourceGateId, destinationGateId)) {
+    if (isAnyGateLinked(sourceGateId, destinationGateId)) {
       revert SmartGate_GateAlreadyLinked(sourceGateId, destinationGateId);
     }
 
@@ -204,6 +204,12 @@ contract SmartGateSystem is EveSystem, AccessModified {
     return true;
   }
 
+  /**
+   * @notice view function to check if the source gate is linked to the destination gate
+   * @param sourceGateId is the smartObjectId of the source gate
+   * @param destinationGateId is the smartObjectId of the destination gate
+   * @return true if the source gate is linked to the destination gate
+   */
   function isGateLinked(uint256 sourceGateId, uint256 destinationGateId) public view returns (bool) {
     SmartGateLinkTableData memory smartGateLinkTableData = SmartGateLinkTable.get(sourceGateId);
     bool isLinked = smartGateLinkTableData.isLinked && smartGateLinkTableData.destinationGateId == destinationGateId;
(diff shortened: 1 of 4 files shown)


mud-contracts/world/test/smart-gate/SmartGateTest.t.sol
```diff
@@ -194,6 +195,76 @@ contract SmartGateTest is MudTest {
...
+    vm.expectRevert(
+      abi.encodeWithSelector(SmartGateSystem.SmartGate_GateAlreadyLinked.selector, smartGateB, smartGateA)
+    );
+    smartGate.linkSmartGates(smartGateB, smartGateA);
+
+    //revert when link BC because B its already linked
+    vm.expectRevert(
+      abi.encodeWithSelector(SmartGateSystem.SmartGate_GateAlreadyLinked.selector, smartGateB, smartGateC)
+    );
+    smartGate.linkSmartGates(smartGateB, smartGateC);
```
vayan, line 229: we should allow A->B, B->C link right?

mud-contracts/world/src/modules/smart-gate/systems/SmartGateSystem.sol
```diff
@@ -121,7 +121,7 @@ contract SmartGateSystem is EveSystem, AccessModified {
    * @param destinationGateId is the smartObjectId of the destination gate
    */
   function linkSmartGates(uint256 sourceGateId, uint256 destinationGateId) public onlyAdminOrObjectOwner(sourceGateId) {
-    if (isGateLinked(sourceGateId, destinationGateId)) {
+    if (isAnyGateLinked(sourceGateId, destinationGateId)) {
```
vayan, line 124: the function should be more like`isGateDestinationLinked`

```sol
function isGateDestinationLinked(uint256 sourceGateId) public view returns (bool) {
  SmartGateLinkTableData memory smartGateLinkTableData = SmartGateLinkTable.get(sourceGateId);

  return smartGateLinkTableData.isLinked;
}
```

because we allow a gate to have multiple sources right? But only one destination

mud-contracts/world/test/smart-gate/SmartGateTest.t.sol
0xxlegolas: Yes I thought about this but If we allow destination then it will break, and this is what happened in that video

A -> B  then we have AB true , BA true
allowing destination
B > C  then we have BC true, CB is true

When we unlink BC, in table it finds all references for B and makes isLinked `false`
BC false 
BA false 

because table has one primary key.

0xxlegolas: So they have to first link A -> B jump and then unlink AB then link BC then jump

mud-contracts/world/src/modules/smart-gate/systems/SmartGateSystem.sol
0xxlegolas: same as above

ccp-raudur: If either the source or the destination gate is linked you shouldn't be able to link it together. So it is correct in the new PR. Gates can't be connected to multiple sources. You can only link A <-> B. We might change that at some point in time.

vayan: if we don't allow A -> B, B -> C chain link then it's ok!

vayan (approved):

mud-contracts/world/test/smart-gate/SmartGateTest.t.sol
ccp-raudur: @vayan for now it is just A <-> B and then C <-> D

ccp-raudur (approved):

PR #363, opened 2025-01-10
improvement: update the sof-v2 to use uint256 IDs for entities

fc-tezza:
A new PR for the SOF V2 to use `uint256` IDs for entities instead of the proposed Id format. This change will make integration with the existing infrastructure much easier.

New Tag structure -

To accommodate this change the Tag structure has undergone an update as well. There are now 3 types of tags: PROPERTY, ENTITY RELATION, and RESOURCE RELATION tags.
**Property Tags** - these are tags to add some property to an entity, the SOF uses these to track the `CLASS` and `OBJECT` properties of an Entity.
**Entity Relation Tags** - these tags are used to track some relationship between Entities. The SOF uses these tags to track the INHERITANCE relationship between OBJECTs and CLASSes
**Resource Relation Tags**- these tag can be used to track some relationship between an Entity and a MUD Resource. The SOF uses these tags to track which MUD Systems are associated via COMPOSITION to an Entity for the purpose of enforcing functionality scope.

All Tags can now have a value assigned to them when they are tagged to an Entity.

More details can be found [here](https://github.com/projectawakening/world-chain-contracts/blob/ccp-tezza/sof-v2-entity-tag-update/mud-contracts/smart-object-framework-v2/README.md) .


0xxlegolas (approved):
LGTM

PR #364, opened 2025-01-13
fuel and smart gate bug fixes

0xxlegolas:
- add additional checks for fuel update to fix when there is a single unit of fuel 
- add additional checks for smart gate ownership check


mud-contracts/world/src/modules/smart-deployable/systems/SmartDeployableSystem.sol
```diff
@@ -150,10 +150,10 @@ contract SmartDeployableSystem is AccessModified, EveSystem, SmartDeployableErro
     }
     _updateFuel(entityId);
     uint256 currentFuel = DeployableFuelBalance.getFuelAmount(entityId);
-    if (currentFuel < 1) revert SmartDeployable_NoFuel(entityId);
+    if (currentFuel < ONE_UNIT_IN_WEI) revert SmartDeployable_NoFuel(entityId);
     DeployableFuelBalance.setFuelAmount(entityId, currentFuel - ONE_UNIT_IN_WEI); //forces it to tick
     _setDeployableState(entityId, previousState, State.ONLINE);
-    DeployableFuelBalance.setLastUpdatedAt(entityId, block.timestamp);
```
vayan, line 156: we don't need this anymore?

vayan (approved):

mud-contracts/world/src/modules/smart-deployable/systems/SmartDeployableSystem.sol
0xxlegolas: no `updatefuel` does this.

vayan (approved):

mud-contracts/world/src/modules/smart-gate/systems/SmartGateSystem.sol
```diff
@@ -134,6 +154,10 @@ contract SmartGateSystem is EveSystem, AccessModified {
       revert SmartGate_NotWithtinRange(sourceGateId, destinationGateId);
     }
 
+    //Delete the existing records for the source and destination gate before creating a new link to avoid replacing the record
+    _deleteExistingLink(sourceGateId);
+    _deleteExistingLink(destinationGateId);
```
tinkrtailor, line 160: Do we have to call this twice if a single call deletes both source and destination link?

mud-contracts/world/src/modules/smart-gate/systems/SmartGateSystem.sol
```diff
@@ -262,6 +286,22 @@ contract SmartGateSystem is EveSystem, AccessModified {
     return distanceSquaredMeters <= (maxDistance * maxDistance);
   }
 
+  /**
+   * @notice delete the existing link between the source and destination gate
+   * @param sourceGateId is the smartObjectId of the source gate
+   */
+  function _deleteExistingLink(uint256 sourceGateId) internal {
```
tinkrtailor, line 294: How is this different from unlink? should we be doing this logic in there? 
`if (!linkData.isLinked) {` indicates that we have an unlinked gate with dangling links that need to be deleted

tinkrtailor (approved):
A few comments regarding this impl. But lgtm

mud-contracts/world/src/modules/smart-gate/systems/SmartGateSystem.sol
0xxlegolas: In that case `_deleteExistingLink` will have duplicate code. 

eg: 
1. SourceGateId can have 2 links. So 2 record deletion 
2. destinationGateId can have 2 links. So 2 record deletion 

So created a common function `_deleteExistingLink` and calling this twice instead of writing the same code.

mud-contracts/world/src/modules/smart-gate/systems/SmartGateSystem.sol
0xxlegolas: yes you are right. Its ideal to delete the dangling ref by deleting the record during unlink. But
The external services are subscribed to `__lastUpdatedBlockNumber` which is a column automatically updated when the record is updated.  Based on the change in the `__lastUpdatedBlockNumber` they inform the game that the gates are unlinked. 

So if we delete the record, then there is no `__lastUpdatedBlockNumber` for that record. So there is no way they can receive a event to inform the game. So when they link we leave the record in the table by marking `isLinked` false so they get updates. 

Later if there is a new link with anyone of the existing gates, then we delete the record.

0xxlegolas: I will add some comments and change the description for more clarity

tinkrtailor (approved):

tinkrtailor (approved):

PR #388, opened 2025-01-30
remove manual execution of scripts during deployment

0xxlegolas:
- added create eve token and setup access control as part of the deploy script in docker 
- sourcing all configurable variables through .env file to docker. 
- adding a new command for world upgrade


build/package/deploy-in-docker.sh
```diff
@@ -1,6 +1,13 @@
 #!/bin/sh
 set -eou pipefail
 
+# Define the log file
+LOG_FILE="./logfile.log"
+mkdir -p logs
+
+# Ensure the log file is copied to the logs folder on exit
+trap 'cp $LOG_FILE "logs/logfile-$(date +%Y%m%d_%H%M%S).log"' EXIT
```
ccp-bofai: I think 'logfile' in the name of a file that has the '.log' extension is probably not needed. Might be more useful to replace it with 'deploy-in-docker' so you can tell at a glance which log file it is. However it might be more practical to have the date time at the start of the file name so you have a better idea of sequence, if this directory contains other log files.

0xxlegolas: fixed in [4770e2e](https://github.com/projectawakening/world-chain-contracts/pull/388/commits/4770e2e3b59281edccb2f08e9c73c214edd1a11a)

ccp-bofai (approved):
Looks good to me

vayan (approved):

tinkrtailor (approved):

PR #417, opened 2025-02-10
fix to verify all delegation flows

0xxlegolas:
- added puppet delegation verification to the forwarder delegation verification to support both meta transaction and regular transaction flows for puppet contracts


vayan (approved):

tinkrtailor:
For context and solidity noobs like me, what is a puppet pattern?

tinkrtailor (approved):

0xxlegolas:
@tinkrtailor its a proxy to forward calls to a system in the MUD world. It delegates the call

PR #428, opened 2025-02-11
improvement: update sof-v2 to be able to handle class/object creation safely from class scoped systems

fc-tezza:
This PR paves the way for the `initialize()` pattern allowing us to bundle System configuration and create classes and objects from our various top level Systems (SmartStorageUnitSystem, SmartCharacterSystem, SmartTurretSystem, SmartGateSystem) as well as create the necessary class or object for for items from InventorySystem and EphemeralInventorySystem.

This includes a new function called `scopedRegisterClass` for EntitySystem which is only callable from specifically defined Systems, as well as changes to the existing `registerClass` and `instantiate` function interfaces to accommodate auto-creation of access roles for these entities upon their creation.

Logic has been added to `deleteClass` and `deleteObject` to remove these auto-generated roles when an entity is deleted.

In addition to this the `scoped` versions of RoleManagementSystem have been relaxed to only check for EntitySystem or Class Scoped System origin. This change allows for a more flexible usage pattern. After full integration we will review the permission profile of each calling function and firm-up any extra access restrictions where needed.  As it stands now though, it should be safe as long as we assume that our logic in the calling Systems (EntitySystem, SSU, SmartCharacter, Inventory, etc...) is correct.

All uint tests updated to accommodate these new interfaces and flows, as well as extensive access configuration testing in `SOFAccessSystem.t.sol`


mud-contracts/smart-object-framework-v2/src/namespaces/evefrontier/systems/entity-system/EntitySystem.sol
```diff
@@ -314,4 +330,50 @@ contract EntitySystem is IEntitySystem, SmartObjectFramework {
...
+
+  function _registerClass(uint256 classId, address accessRoleMember, ResourceId[] memory scopedSystemIds) private {
+    if (classId == uint256(0)) {
+      revert Entity_InvalidEntityId(classId);
+    }
+    if (Entity.getExists(classId)) {
+      revert Entity_EntityAlreadyExists(classId);
+    }
+
+    bytes32 classAccessRole = keccak256(abi.encodePacked("ACCESS_ROLE", classId));
```
0xxlegolas, line 307: CLASS_ ACCESS_ROLE instead of ""ACCESS_ROLE"

0xxlegolas (approved):

mud-contracts/smart-object-framework-v2/src/namespaces/sofaccess/systems/sof-access-system/SOFAccessSystem.sol
```diff
@@ -153,21 +142,67 @@ contract SOFAccessSystem is ISOFAccessSystem, SmartObjectFramework {
...
     if (callCount > 1) {
-      // if not a direct entrypoint call to TagSystem, (instead a subsequent internal call), then we only allow EntitySystem.sol as the target System of this function
-      (, , address msgSender, ) = IWorldWithContext(_world()).getWorldCallContext();
+      // system-to-system call case
       ResourceId callingSystemId = SystemRegistry.get(msgSender);
-      if (callingSystemId.unwrap() != EntitySystemUtils.entitySystemId().unwrap()) {
-        // for TagSystem if an internal call is not from EntitySytem or a Class scoped system then we reject the call
-        revert SOFAccess_SystemAccessDenied(entityId, msgSender);
+      if (
+        callingSystemId.unwrap() == InventoryUtils.ephemeralInventorySystemId().unwrap() ||
```
0xxlegolas: Is it a good idea to hardcode all the systems from world in sof access ? 
Should we store in a table and read it perhaps ?

Kooshaba: i agree we shouldn't have this hardcoded, feels like a mistake waiting to happen. if it's in a table then it can live alongside all of the other setup so it is easier to remember to change.

Kooshaba (changes requested):

mud-contracts/smart-object-framework-v2/src/namespaces/sofaccess/systems/sof-access-system/SOFAccessSystem.sol
```diff
@@ -20,6 +20,11 @@ import { IWorldWithContext } from "../../../../IWorldWithContext.sol";
 
 import { IEntitySystem } from "../../../evefrontier/interfaces/IEntitySystem.sol";
 import { Utils as EntitySystemUtils } from "../../../evefrontier/systems/entity-system/Utils.sol";
+import { InventoryUtils } from "@eveworld/world-v2/src/namespaces/evefrontier/systems/inventory/InventoryUtils.sol";
+import { SmartStorageUnitUtils } from "@eveworld/world-v2/src/namespaces/evefrontier/systems/smart-storage-unit/SmartStorageUnitUtils.sol";
+import { SmartCharacterUtils } from "@eveworld/world-v2/src/namespaces/evefrontier/systems/smart-character/SmartCharacterUtils.sol";
+import { SmartGateUtils } from "@eveworld/world-v2/src/namespaces/evefrontier/systems/smart-gate/SmartGateUtils.sol";
+import { SmartTurretUtils } from "@eveworld/world-v2/src/namespaces/evefrontier/systems/smart-turret/SmartTurretUtils.sol";
```
Kooshaba: instead of importing the utils we can use the generated libraries now and use `toResourceId()`

mud-contracts/smart-object-framework-v2/src/namespaces/evefrontier/systems/entity-system/EntitySystem.sol
fc-tezza: I was using the string "ACCESS_ROLE" for both the `classId` and `objectId` cases since this then simplifies the interface for those who would want to interact with it.. a standardized way to call using "ACCESS_ROLE" + `entityId`. This is safe because both `classIds` and `objectIds` are guaranteed unique in the `Entity` Table and so will always produce different hashes regardless of the descriptor string.

mud-contracts/smart-object-framework-v2/src/namespaces/sofaccess/systems/sof-access-system/SOFAccessSystem.sol
fc-tezza: Yes, I am planning to convert to using the system libs pattern in the next PR

mud-contracts/smart-object-framework-v2/src/namespaces/sofaccess/systems/sof-access-system/SOFAccessSystem.sol
fc-tezza: This is true and I considered it.. but the trade-off here is that if we make this adjustable by Table value, it adds overhead in terms of needing an extra config function to set the Table values, and the need for another Table in the MUD config. 

We can change to a Table valued pattern, but I opted to not add that for the first cut in the name of simplicity. Particularly, since it may be that some of these access functions are not stable yet and may change moving forwards.

My plan was to do one last round of clean-up for the SOF after we are done with all of the integration changes. I can add this to the clean up list for that PR.

mud-contracts/smart-object-framework-v2/src/namespaces/sofaccess/systems/sof-access-system/SOFAccessSystem.sol
fc-tezza: But does this need to change to integrate with the latest world v2 branch? Did we remove Util files there already?

0xxlegolas: No I suppose it should not affect , it can be done in the next PR

Kooshaba (approved):

PR #429, opened 2025-02-13
chore: update TypeScript configuration to support modern esm

0xxlegolas:
- supports modern esm syntax


Kooshaba (approved):
works for me 👍🏼

fc-tezza (approved):

PR #451, opened 2025-02-24
top level initialisation

0xxlegolas:
- top level initialisation for class registration
- top level access configuration
- fix inventory test out of gas error
- clean up builds


mud-contracts/world-v2/test/EveTest.sol
```diff
@@ -52,6 +52,8 @@ import { smartTurretSystem } from "../src/namespaces/evefrontier/codegen/systems
 import { SmartGateSystem } from "../src/namespaces/evefrontier/systems/smart-gate/SmartGateSystem.sol";
 import { smartGateSystem } from "../src/namespaces/evefrontier/codegen/systems/SmartGateSystemLib.sol";
 
+import { Initialize } from "../src/namespaces/evefrontier/codegen/index.sol";
+
 abstract contract EveTest is Test {
```
Kooshaba, line 57: we should use EveSystem here as well instead of having duplicate logic

Kooshaba (changes requested):

Kooshaba (approved):

fc-tezza (commented):
https://github.com/projectawakening/world-chain-contracts/commit/3828e6e57c4f06ea1057916de31e8095311b7b53
This patch has been merge so that we can continue progress on this initialization PR.

Please note: a few discussed changes
- Script needed for world configurations in the SOF - https://github.com/projectawakening/world-chain-contracts/pull/456/files/7c33f1cd76f3c25ba7a80a12d2d2f7079c8f963d#r1976777761
- Script needed to add EveSystem to the CallAccess table - https://github.com/projectawakening/world-chain-contracts/blob/3828e6e57c4f06ea1057916de31e8095311b7b53/mud-contracts/smart-object-framework-v2/scripts/SetSOFCallAccess.s.sol#L37 https://github.com/projectawakening/world-chain-contracts/blob/3828e6e57c4f06ea1057916de31e8095311b7b53/mud-contracts/smart-object-framework-v2/scripts/SetSOFCallAccess.s.sol#L45
- removal of `world()` in `EveSystem.sol` - it is no longer needed now that we have MUD system libs
- Possibly using interfaces as imports in EveSystem.sol instead of full contracts.
- `EveSystem.initialise()` should use `EntitySystem.scopedRegisterClass` instead of `registerClass`
- `EveSystem.initialise()` - no need to create a role, this is automatically done in the `_registerClass` logic
- We should consider making the EveSystem functionality only callable by the namespace owner to avoid any possible security complications during World configuration.

fc-tezza (approved):

PR #474, opened 2025-03-04
refactor eve tests

0xxlegolas:
- runs test against the smart object framework by creating a anvil snapshot and using it 
- world configuration scripts for smart object framework classes and access 
- removed EVE test
- refactored all tests to work against the new flow


fc-tezza (approved):
All looks good to me

PR #492, opened 2025-03-11
chore: test script tweak

Kooshaba:
1. `set -e` in bash script makes sure there are no silent failures
2. `.env` was invalid for me because of quotes


0xxlegolas (approved):
LGTM

mud-contracts/world-v2/.env
```diff
@@ -13,7 +13,7 @@ RPC_URL=http://127.0.0.1:8545
 WORLD_ADDRESS=
 
 
-ERC20_TOKEN_NAME=Test TOKEN
+ERC20_TOKEN_NAME="Test TOKEN"
```
0xxlegolas, line 16: good catch

PR #496, opened 2025-03-19
erc721 removal and inventory cleanup

0xxlegolas:
**### Change Log:**

**ERC721 related -**

- Remove all ERC721 code from world-v2
- Remove all references to ERC721 from Deployable, SmartCharacter, and scripts
- Remove StaticData Table, Static Data System, and all Static data reference logic
- Remove all ERC721 support tables from [mud.co](http://mud.co/)nfig
- Created an Ownership system which tracks ultimate beneficiary account ownership for both singleton and non-singleton smart objects. As well as tracking aggregate quantities of objects in the world and aggregate quantities for each individual account.
- Only items in inventory can be transferred.
- Only items in inventory ore singleton or non-singleton direct calls to deposit/withdrawal (mint/burn ownership records, call from our "interact" systems just transfer ownership
- Added ownership related tables to the mud.config
- Add new ownership tracking logic to SmartCharacter, Deployable, Inventory, and EphemeralInventory for managing account ownership tracking
- new character/deployables create will "assign" new ownership to things-in-space
- `removeCharacter`, `unanchor`, and `destoryDeployable` will "removel" ownership from things-in-space
- direct calls by the server to Inventory/Ephemeral deposit/withdrawal will also assign/remove ownership of items because this action is also like adding/removing ownership tracking of items to/from the game simulation
- any inventory calls that originate from our "interact" contracts (not a direct call) will transfer holdership from one inventory to another, and transfer ownership from one inventory owner to another 

**SmartCharacter related -**

- Update param naming convention in SmartCharacter to use `smartObjectId` for consistency
- Adjusted Characters Table (remove `characterAddress` since all smartobject-to-account tracking can now be handled the OwnershipSystem)
- Add `removeCharacter` logic to SmartCharacterSystem
- error reporting updates

**Deployable related -**

- Adjusted the `unanchor` and `destroyDeployable` cases to also remove ownership linkage 
- Adjusted `Deployable.createDeploayble` to only allow the NULL state, Not UNANCHORED . This was a previous workaround, but this is unsafe , when you re-anchor you should call anchor not create.
- function name shortening where possible
- error reporting updates
- Inventory version updating when `unanchor` or `destroy` called
- Inventory used capacity reset when `unanchor` or `destroy` called
- smart gate unlink called when `unanchor` or `destroy` called
- `owner` added to anchor params, to track ownership upon re-anchoring

**Fuel related -** 

- add strict parameter range restrictions to prevent any overflow/underflow cases in Fuel logic

**Location related -** 
- remove `Coord` and `WorldLocation` types in favor of using the LocationData table struct itself

**EntityRecords related -**

- Add a Tenant table for tracking the acceptable tenant value for this World for EntityRecords
- Update EntityRecord with `tenantId` for full game data tracking
- Add EntityRecord sanity checks to Inventory/Ephemeral Item create flows, as well as to Deployable creation (in SmartAssembly) and Smart Character creation.
- function name shortening where possible
- make EntityRecord fields consistent everywhere in the whole codebase

**AccessSystem related -**

- Made AccessSystem consistent in terms of param and function naming (use smartObjectId everywhere).
- Added several Access logic rule for configuration
- Update AccessSystem to use new `Ownership.owner()` getter for `isOwner` checks
- Re-implement role access rules for Inventory/EphemeralInteract in AccessSystem
- adjusted the access for `EntitySystem.deleteClass` to NO ACCESS, deleting a class currently leaves a bunch of orphaned object data with unhandled data corruption, so we prohibit it for now.
- Implement Ownership access rule with scope enforcement for functions which can have ephemeral objects as an input - by using ObjectByEphemeral lookups
- Added special case access for entity record - for SmartCharacter - only admin can set name, and set metadata
- Add EntityRecord parameter sanity checks to Inventory, EphemeralInventory and Deployable create functions
- Add `CallAccess` rules to Access System (standardized way for system-to-system call access)

**Inventory related -**

- Updated Inventory and EphemeralInventory item ID generation logic (I.e. only create classId EntityRecords and Entity entries when they don’t exist)
- change `inventoryItemId` naming to `itemObjectId` for consistency in `mud.config` and elsewhere
- Added `NULL` to  the deployable state checks for inventory actions (this will allow non-deployables to also have inventory, since no created Deployable has state NULL) this is to accommodate adding inventory to rifts which are Assemblies (but not Deployable), and will have inventory
- Reworked Inventory/Interact role logic - no more role assignment in the capacity setting logic. Now the SmartObject Owner (as per the OwnershipSystem logic) can set InventoryInteract roles for who can interact with functionality.
- Updated the role setting logic to use these InventoryInteract specific roles and the functions themselves now check those specific roles for access.
- Moved `setEphemeralCapacity` to the InventorySystem.. since this is global capacity setter at the SmartObject level, whereas all other EphemeralInventory functionality is per ephemeral inventory
- Reworked Inventory/Ephemeral deposit flows to better match the withdrawal logic flows
- Rename InventoryInteract to `EphemeralInteract` (as in... this contract is for interacting with Ephemeral inventories from an Object's primary inventory)
- Created an new contract called `InventoryInteract` (for explicitly interacting with other smart object inventories from a primary inventory)
- `InventoryInteract` has exactly same pattern as `EphemeralInteract` except this handles transfers from one smart object inventory to another object inventory (instead of a smart object inventory to an associated ephemeral inventory)
- Added another transfer tracking table called `InventoryItemTransfer` to track object-to-object inventory transfers
- Renamed `ItemTransfer` table to `EphemeralItemTransfer`
- function name shortening where possible
- error reporting updates
- decoupled item data versioning from deployable state, it now only relies on the inventory `version` for stale data tracking

**SmartGate related -**
- Update link unlink logic
- Add auto-unlinking in Deployable when state changes to `UNANCHORED` or `DESTROYED`

**Parameter/Types related -**

- Rename CreateAndAnchorDeployableParams to `CreateAndAnchorParams`
- Remove `SmartObjectData` type from CreateAndAnchorParams
- rename InventoryItem to `CreateInventoryItemParams` and TransferItem to `InventoryItemParams`.
- Use `InventoryItemParams` consistently everywhere (besides the create functions)
- remove the `owner` param from both CreateInventoryItemParams and InventoryItemParams
- Change name of EntityRecordData to `EntityRecordParams` and EntityMetadata to `EntityMetadataParams`, to avoid conflict with the similarly named Table struct and consistency.

**General -**

- Remove SOFAccess Utils & Remove InventoryUtils
- Remove `common-constants` dependency in world-v2 package.json
- Rebase with the latest develop branch and init pattern changes
- Rename ConfigSof to `Config`, since this script configures the EveSystem setup and World access also (not just SOF)
- Add the `OwnershipSystem` to EveSystem for class initializations
- Added `Tenant` setting to the `Config` script and the .env
- Update `Config` script to reflect the latest changes
- Configure Inventory/EphemeralInventory/Ownership CallAccess in Config script
- Update naming and access matrix in EveSystem
- Add role based access control configurations for the OwnershipSystem and Inventory/EpemeralInteract updates to EveSystem
- `tenetId` added to classId generation logic, to prevent cross-chain asset usage
- `EntityRecord` creation added to the class `initialize()` flow - this will be updated in the future to be more robust and lees reliant on the game data directly


mud-contracts/world-v2/src/namespaces/evefrontier/systems/EveSystem.sol
```diff
@@ -510,11 +614,16 @@ contract EveSystem is IEveSystem, SmartObjectFramework {
...
    * @param systemIds The system ids that belong to the class
    */
-  function initialize(uint256 typeId, ResourceId[] memory systemIds) internal returns (uint256) {
+  function initialize(uint256 typeId, uint256 volume, ResourceId[] memory systemIds) internal returns (uint256) {
     if (typeId == 0) revert("Invalid typeId");
-    uint256 classId = uint256(keccak256(abi.encodePacked(typeId)));
+    uint256 classId = uint256(keccak256(abi.encodePacked(Tenant.get(), typeId)));
     entitySystem.scopedRegisterClass(classId, _callMsgSender(1), systemIds);
-
+    entityRecordSystem.createRecord(classId, EntityRecordParams({
```
0xxlegolas: what is the issue in creating a entity record when a ssu creation flow starts from the game ? 
Why should this be part of deployment process ?

mud-contracts/world-v2/src/namespaces/evefrontier/systems/EveSystem.sol
```diff
@@ -108,89 +121,135 @@ contract EveSystem is IEveSystem, SmartObjectFramework {
...
-  // Configure access for StaticDataSystem
-  function configureStaticDataAccess() public {
-    bytes4[2] memory onlyAdminSelectors = [StaticDataSystem.setCid.selector, StaticDataSystem.setBaseURI.selector];
+    bytes4[2] memory entityRecordOnlyAdminForCharactersOtherwiseAlsoOwnerSelectors = [
+      EntityRecordSystem.createMetadata.selector,
+      EntityRecordSystem.setName.selector
+    ];
 
-    for (uint256 i = 0; i < onlyAdminSelectors.length; i++) {
+    for (uint256 i = 0; i < entityRecordOnlyAdminForCharactersOtherwiseAlsoOwnerSelectors.length; i++) {
```
0xxlegolas: `entityRecordOnlyAdminForCharactersOtherwiseAlsoOwnerSelectors ` the variable name is too long. Can we stick to length <25 for function names and variable names ?

mud-contracts/world-v2/src/namespaces/evefrontier/systems/deployable/DeployableSystem.sol
```diff
@@ -197,13 +210,22 @@ contract DeployableSystem is SmartObjectFramework {
...
     if (previousState != State.UNANCHORED) {
       revert Deployable_IncorrectState(smartObjectId, previousState);
     }
     _setDeployableState(smartObjectId, previousState, State.ANCHORED);
+    // ascribe ownership tracking of the deployable smart object
+    address currentOwner = ownershipSystem.owner(smartObjectId);
+    if (currentOwner != address(0) && currentOwner != owner) {
+      ownershipSystem.annulFromAccount(smartObjectId, currentOwner);
+      ownershipSystem.ascribeToAccount(smartObjectId, owner);
+    } else if (currentOwner == address(0)) {
```
0xxlegolas: might be best if the cases are swapped as the most used case is `currentOwner == address(0)` so it does not have to go to else in most cases

mud-contracts/world-v2/mud.config.ts
```diff
@@ -31,35 +31,47 @@ export default defineWorld({
...
         },
         /***************************
-         * SMART ASSEMBLY *
+         * OWNERSHIP TABLES *
          ***************************/
-        /**
-         * Used to store the assembly of a smart object
-         */
-        SmartAssembly: {
+        InventoryByItem: {
```
0xxlegolas, line 36:
```suggestion
        InventoryItem: {
```

mud-contracts/world-v2/mud.config.ts
```diff
@@ -31,35 +31,47 @@ export default defineWorld({
...
+          schema: {
+            itemObjectId: "uint256",
+            inventoryObjectId: "uint256",
+          },
+          key: ["itemObjectId"],
+        },
+        OwnershipByObject: {
```
0xxlegolas, line 43:
```suggestion
        Owner: {
```

mud-contracts/world-v2/mud.config.ts
```diff
@@ -72,134 +84,42 @@ export default defineWorld({
...
-          },
-          key: ["characterAddress"],
+          key: ["smartObjectId"],
         },
-
-        /*************************
-         * ERC721 PUPPET MODULE *
-         ************************/
-        Balances: {
+        CharactersByAccount: {
```
0xxlegolas, line 112:
```suggestion
        Characters: {
```

mud-contracts/world-v2/mud.config.ts
```diff
@@ -272,33 +181,51 @@ export default defineWorld({
...
             index: "uint256",
-            stateUpdate: "uint256",
+            version: "uint256",
+          },
+          key: ["smartObjectId", "itemObjectId"],
+        },
+        /**
+         * Used to signal the transfer details when a item is exchanged between its primary inventory and another smart object's primary inventory
+         */
+        ObjectItemTransfer: {
```
0xxlegolas:
```suggestion
        InventoryItemTransfer: {
```

mud-contracts/world-v2/mud.config.ts
```diff
@@ -372,7 +303,6 @@ export default defineWorld({
           },
           key: ["smartObjectId"],
         },
-
         SmartGateLink: {
```
0xxlegolas, line 308: Made couple of comments on the table names to keep it simple and intuitive. Let me know if it works

mud-contracts/world-v2/mud.config.ts
fc-tezza: these are different tables, storing different things, we already have an `InventoryItem` table

fc-tezza: I'm using the CharactersByAccount pattern for naming for reverse lookup tables

mud-contracts/world-v2/mud.config.ts
fc-tezza: I'm using the CharactersByAccount pattern for naming for reverse lookup tables

mud-contracts/world-v2/mud.config.ts
fc-tezza: We already have a `Characters` table.. `CharactersByAccount `is the reverse lookup table

mud-contracts/world-v2/mud.config.ts
fc-tezza: Sounds good

mud-contracts/world-v2/src/namespaces/evefrontier/systems/EveSystem.sol
fc-tezza: yeah, it's hard to be descriptive with shorter versions, but I agree these names are too long!

mud-contracts/world-v2/src/namespaces/evefrontier/systems/EveSystem.sol
fc-tezza: This is a stop gap solution until we settle exactly how we are going to get class EntityRecord data. We could use the inventory pattern of generation just-in-time. 

But I also viewed that as a temporary (flawed) solution since it is not particularly robust. it relies on the fact that the game servers will give us accurate data in every call, which seems pretty brittle.

Ultimately, I think the class data pre-population flow combined with some decoupling of IDs from hard reliance on game data is probably the best long term solution.

0xxlegolas: > But I also viewed that as a temporary (flawed) solution since it is not particularly robust. it relies on the fact that the game servers will give us accurate data in every call, which seems pretty brittle.

The same issue could happen during deployment too, right? 

> Ultimately, I think the class data pre-population flow combined with some decoupling of IDs from hard reliance on game data is probably the best long term solution.
agreed

mud-contracts/world-v2/mud.config.ts
0xxlegolas: got it

mud-contracts/smart-object-framework-v2/src/namespaces/sofaccess/systems/sof-access-system/SOFAccessSystem.sol
```diff
@@ -257,6 +255,21 @@ contract SOFAccessSystem is ISOFAccessSystem, SmartObjectFramework {
     revert SOFAccess_AccessDenied(entityId, msgSender);
   }
 
+    /**
+   * @notice Blocks all calls
+   * @param entityId The ID of the entity (class or object) to check
+   * @param targetCallData The calldata of the target function
+   * @dev Handles calls to EntitySystem.deleteClass, currently there are too many un-resolved data dependencies to allow for any access to this function
+   */
+  function noAllowances(uint256 entityId, bytes memory targetCallData) public view {
```
0xxlegolas, line 264: Just curious—if some functions aren’t meant to be accessed anymore, is there a reason to keep them around? Wondering if it makes sense to keep obsolete code.

mud-contracts/world-v2/mud.config.ts
0xxlegolas: Then `ItemToInventoryMap` ?

mud-contracts/world-v2/mud.config.ts
0xxlegolas: Does this table also track ephemeral to ephemeral transfer of items ?

mud-contracts/world-v2/mud.config.ts
```diff
@@ -309,49 +236,54 @@ export default defineWorld({
...
+            exists: "bool",
             quantity: "uint256",
             index: "uint256",
-            stateUpdate: "uint256",
+            version: "uint256",
           },
-          key: ["smartObjectId", "inventoryItemId", "ephemeralInvOwner"],
+          key: ["smartObjectId", "ephemeralOwner", "itemObjectId"],
+        },
+        ObjectByEphemeral: {
```
0xxlegolas: What’s the purpose of the ObjectByEphemeral table? I’m having trouble understanding its use—since the key is ephemeralSmartObjectId, it seems like the user would already need both smartObjectId and ephemeralOwner to construct it. If they already have those values, is there really a need to look it up in a table?

mud-contracts/world-v2/mud.config.ts
```diff
@@ -309,49 +236,54 @@ export default defineWorld({
...
+            ephemeralOwner: "address",
+          },
+          key: ["ephemeralSmartObjectId"],
         },
         /**
-         * Used to store the transfer details when a item is exchanged
+         * Used to signal the transfer details when a item is exchanged between it's primary inventory and an associated ephemeral inventory or two ehpemeral inventories
          */
-        ItemTransferOffchain: {
+        ItemTransfer: {
```
0xxlegolas: Now that we have two types of transfers—between different inventories and within the same one—would it make sense to rename them to `IntraInventoryEphemeralTransfer` and `InterInventoryEphemeralTransfer`? Might help make things more intuitive for integrations

mud-contracts/world-v2/src/namespaces/evefrontier/systems/deployable/DeployableSystem.sol
```diff
@@ -106,20 +105,22 @@ contract DeployableSystem is SmartObjectFramework {
...
     }
 
-    if (previousState == State.NULL) {
-      address erc721Address = DeployableToken.getErc721Address();
-      IERC721Mintable(erc721Address).mint(smartObjectData.owner, smartObjectId);
-
-      staticDataSystem.setCid(smartObjectId, smartObjectData.tokenURI);
+    // TODO: the following is a candidate for hook logic
+    // set the initial inventory data version to 1
+    if (Inventory.getVersion(smartObjectId) == 0) {
```
0xxlegolas: Shouldn't we revert if its not 0 ?

mud-contracts/world-v2/src/namespaces/evefrontier/systems/deployable/DeployableSystem.sol
```diff
@@ -197,13 +210,22 @@ contract DeployableSystem is SmartObjectFramework {
    */
   function anchor(
     uint256 smartObjectId,
+    address owner,
     LocationData memory locationData
   ) public onlyActive context access(smartObjectId) scope(smartObjectId) {
     State previousState = DeployableState.getCurrentState(smartObjectId);
     if (previousState != State.UNANCHORED) {
```
0xxlegolas, line 235: Should we also include || previousState == State.UNANCHORED here? Since a deployable can go from unanchored to anchored ? 

May be this was the reason the above condition was added in createDeployable

mud-contracts/world-v2/src/namespaces/evefrontier/systems/inventory/EphemeralInventorySystem.sol
```diff
@@ -1,346 +1,366 @@
...
+    // Ensure the smartObjectId is not itself an ephemeral object
+    if (OwnershipByObject.get(smartObjectId) == address(0)) {
+      revert EphemeralInventory_InvalidSmartObjectId(smartObjectId);
+    }
+    
+    // Validate state (uses the associated smart object's state). This entails that the associated smart object exists and is anchored or online.
     {
       State currentState = DeployableState.getCurrentState(smartObjectId);
-      if (currentState != State.ONLINE) {
+      if (!(currentState == State.NULL || currentState == State.ONLINE)) {
```
0xxlegolas, line 116: why `NULL` state is included ?

0xxlegolas: I saw the comment its for the inventories not attached with deployables

mud-contracts/world-v2/mud.config.ts
0xxlegolas: Sorry, noticed ephemeral to ephemeral transfer is also happening, so we can leave it as it is.

mud-contracts/world-v2/mud.config.ts
0xxlegolas: I saw itemTransfer table is also doing something similar so can we rename as `IntraInventoryTransfer`

mud-contracts/world-v2/src/namespaces/evefrontier/systems/smart-assembly/SmartAssemblySystem.sol
```diff
@@ -1,82 +1,107 @@
...
+    string memory assemblyType
+  ) public context access(smartObjectId) scope(smartObjectId) {
+    if ((keccak256(abi.encodePacked(assemblyType)) == keccak256(abi.encodePacked("")))) {
+      revert SmartAssembly_TypeCannotBeEmpty(smartObjectId);
     }
 
-    uint256 smartAssemblyEnumId = SmartAssembly.getSmartAssemblyId(smartObjectId);
-    smartAssemblyEnumId = smartAssemblyEnumId + 1;
+    uint256 assemblyEnumId = SmartAssembly.getAssemblyId(smartObjectId);
+    assemblyEnumId = assemblyEnumId + 1;
```
0xxlegolas: The enum ID should probably be predefined—like SSU = 0, ST = 1, SG = 2. If we’re incrementing it with every createAssembly, which now gets called during every createAndAnchor, then we’ll end up incrementing the ID for each SSU. Also, I noticed that createSmartAssembly is only being used for deployables, not for gates or turrets

mud-contracts/smart-object-framework-v2/src/namespaces/sofaccess/systems/sof-access-system/SOFAccessSystem.sol
fc-tezza: `deleteClass` is prohibited for now until we can find a graceful solution for data dependency issues

mud-contracts/world-v2/mud.config.ts
fc-tezza: Ephemeral inventories are assigned an ephemeral object id which is generated combining their lookup keys (it is a unique ID `keccack256(inventorySmartObjectId, ephemeralOwner)`

This `ephemeralObjectId` populated in `InventoryByItem.set(itemObjectId, ephemeralInventoryId)` when an item is assigned to an ephemeral inventory.

For lookup purposes we also then have `ObjectByEphemeral` which is another reverse lookup table which allows us to input the `ephermalObjectId `to lookup the associated primary inventory and the owner.

With this we can find both the primary inventory smart object Id (the ID of the SSU or other deployable) and the owner of the ephemeral inventory holding that item without having to know those pieces of dat beforehand.. 

Under this pattern, we only need to input the `itemObjectId` and we can find all of this ownership data.

mud-contracts/world-v2/src/namespaces/evefrontier/systems/deployable/DeployableSystem.sol
fc-tezza: We are checking the UNANCHORED requirement on line 216, just below your comment here.

mud-contracts/world-v2/src/namespaces/evefrontier/systems/deployable/DeployableSystem.sol
fc-tezza: Currently there is no possibility for getting this far in the code with a non-zero version.

There are only two places that version gets bumped - when a deployable is unanchored or destroyed. In both case the state of the deployable will be not-NULL and can never go back to the NULL state.

Because of this the earlier NULL state check would prohibit us from ever getting this far in the code with a non-zero  versioned inventory, because any such deployable would fail the NULL check.

so actually instead, I guess we don't even need a zero check here.. we can just set to 1 regardless because at this point it will ALWAYS be a zero version.

fc-tezza: But this did remind me that not all deployables necessarily have an inventory, so I've updated the code to check and see if the current deployable smartObject has Inventory scoped/tagged to its class and only set a version if an inventory is possible.

mud-contracts/world-v2/src/namespaces/evefrontier/systems/smart-assembly/SmartAssemblySystem.sol
fc-tezza: > The enum ID should probably be predefined—like SSU = 0, ST = 1, SG = 2. If we’re incrementing it with every createAssembly, which now gets called during every createAndAnchor, then we’ll end up incrementing the ID for each SSU.

I think the current logic just assigns 1 every time to each `smartObjectId`, unless that `smartObjectId `has had its assemblyType set before (which shouldn't/doesn't happen in our code).

So I don't think this will do as you mention, but I do think we should review what exactly we want this to do.

If we want to make it static and want it independently trackable from `typeId` data, we could just as easily `unit256(keccak256(assemblyType))` for its uint id. Was there a reason we want it to be an enum?

> Also, I noticed that createSmartAssembly is only being used for deployables, not for gates or turrets

gates and turrets ARE deployables, so they are covered.

mud-contracts/world-v2/mud.config.ts
0xxlegolas: The naming got me confused, Does this work 

```
/**
         * Look up table to find the associated inventory smart object for an ephemeral inventory
         */
        InventoryByEphemeral: {
          schema: {
            ephemeralSmartObjectId: "uint256",
            exists: "bool",
            inventorySmartObjectId: "uint256", // parent container ID
            inventoryOwner: "address", 
          },
          key: ["ephemeralSmartObjectId"],
        },
```

mud-contracts/world-v2/src/namespaces/evefrontier/systems/smart-assembly/SmartAssemblySystem.sol
0xxlegolas: The best way could be defining  like this  :  https://github.com/projectawakening/world-chain-contracts/blob/develop/mud-contracts/world/mud.config.ts#L69

mud-contracts/world-v2/mud.config.ts
fc-tezza: I've opted for EphemeralItemTransfer (for EphemeralInteract) and InventoryItemTransfer (for InventoryInteract)

mud-contracts/world-v2/src/namespaces/evefrontier/systems/EveSystem.sol
fc-tezza: > The same issue could happen during deployment too, right?

Yes but at least there it is only one time and easily track-able. During on-going operational calls would be far harder to monitor and find issues. Regardless, we should update this in the near-ish future.

mud-contracts/world-v2/mud.config.ts
0xxlegolas: This sounds good but I noticed 

`InventoryItemTransfer` does inventory to inventory transfer 
`EphemeralItemTransfer` does inventory to ephemeral, ephemeral to inventory and ephemeral to ephemeral.

mud-contracts/world-v2/src/namespaces/evefrontier/systems/smart-gate/SmartGateSystem.sol
```diff
@@ -52,26 +54,36 @@ contract SmartGateSystem is SmartObjectFramework {
...
       revert SmartGate_SameSourceAndDestination(sourceGateId, destinationGateId);
     }
 
-    //TODO: Check if the state is online for both the gates ??
     if (isWithinRange(sourceGateId, destinationGateId) == false) {
       revert SmartGate_NotWithtinRange(sourceGateId, destinationGateId);
     }
 
-    //Create a 2 way link between the gates
+    // Create a 2 way link between the gates
```
0xxlegolas, line 91: delete links before unlink is missing 

```
    //Delete the existing records for the source and destination gate before creating a new link to avoid replacing the record
    //The invalid records are not deleted during unlink because the external services are subscribed to the unlink events. If the record is deleted then the external services will not be able to notify the game
    _deleteExistingLink(sourceGateId);
    _deleteExistingLink(destinationGateId);
```

mud-contracts/world-v2/mud.config.ts
fc-tezza: yeah, the `Ephemeral` naming just signifies that it handles everything to do with Ephemeral interactions.

mud-contracts/world-v2/src/namespaces/evefrontier/systems/smart-gate/SmartGateSystem.sol
```diff
@@ -113,19 +127,19 @@ contract SmartGateSystem is SmartObjectFramework {
...
-
-    State destinationGateState = DeployableState.getCurrentState(destinationGateId);
-
-    if (sourceGateState != State.ONLINE) {
-      revert DeployableSystem.Deployable_IncorrectState(sourceGateId, sourceGateState);
-    }
-
-    if (destinationGateState != State.ONLINE) {
-      revert DeployableSystem.Deployable_IncorrectState(destinationGateId, destinationGateState);
+    // Check if the gates are online
```
0xxlegolas: Do we really need multiple if-else checks here? Isn’t the main goal just to show which gate isn’t online?

mud-contracts/world-v2/mud.config.ts
fc-tezza: Yeah, sounds good

mud-contracts/world-v2/src/namespaces/evefrontier/systems/smart-assembly/SmartAssemblySystem.sol
fc-tezza: As we discussed, I've just removed the enum number tracking completely since we don't use it anywhere currently.

mud-contracts/world-v2/src/namespaces/evefrontier/systems/smart-gate/SmartGateSystem.sol
fc-tezza: Links are now auto-deleted in the Deployable logic [here](https://github.com/projectawakening/world-chain-contracts/blob/95f1df71d7c6f24855601e0b5fa12da28c4fee54/mud-contracts/world-v2/src/namespaces/evefrontier/systems/deployable/DeployableSystem.sol#L172) and [here](https://github.com/projectawakening/world-chain-contracts/blob/95f1df71d7c6f24855601e0b5fa12da28c4fee54/mud-contracts/world-v2/src/namespaces/evefrontier/systems/deployable/DeployableSystem.sol#L281), or manually [here](https://github.com/projectawakening/world-chain-contracts/blob/95f1df71d7c6f24855601e0b5fa12da28c4fee54/mud-contracts/world-v2/src/namespaces/evefrontier/systems/smart-gate/SmartGateSystem.sol#L96). 
There is no need to have the delete logic in this function now.. if either gate has a link still, [we will revert before make it this far](https://github.com/projectawakening/world-chain-contracts/blob/95f1df71d7c6f24855601e0b5fa12da28c4fee54/mud-contracts/world-v2/src/namespaces/evefrontier/systems/smart-gate/SmartGateSystem.sol#L74) in the logic. And if there were old link data still, we will just overwrite any old link data with new links [here](https://github.com/projectawakening/world-chain-contracts/blob/95f1df71d7c6f24855601e0b5fa12da28c4fee54/mud-contracts/world-v2/src/namespaces/evefrontier/systems/smart-gate/SmartGateSystem.sol#L87).

fc-tezza: The unlink event will still be triggered.. but it will come from the deployable logic now when appropriate state change occurs.

fc-tezza: I've added this back in to satisfy our integration needs. We should explore a better pattern for the integration data needs.

mud-contracts/world-v2/src/namespaces/evefrontier/systems/smart-gate/SmartGateSystem.sol
fc-tezza: I just thought it is useful information to know which gate is not online and blocking the call for error reporting, 

We could collapse this into `source-not-online || destination-not-online` but then we lose the ability to report which gate is the offender in the error.

0xxlegolas: Based on the previous code, if we do something like:
```
if (sourceGateState != State.ONLINE) {  
   // revert: source gate is not online  
}  
  
if (destinationGateState != State.ONLINE) {  
   // revert: destination gate is not online  
}  
```
this should indicate which gate isn’t online, right?

fc-tezza (approved):

mud-contracts/world-v2/src/namespaces/evefrontier/systems/smart-gate/SmartGateSystem.sol
fc-tezza: But that doesn't cover the case of reporting that BOTH of them are offline, which is why it is structured how it is currently.

PR #497, opened 2025-03-24
improvement/refactor and fixes

0xxlegolas:


ccp-tezza:
I've taken a look and merged this back into #496 . All looking great! 

I only made one change for naming -
`assignOwnerToInventory` -> `assignItemToInventory`
`removeOwnerFromInventory` -> `removeItemFromInventory`

These functions are adding/removing item quantity and ownership tracking to inventories, but they are not actually adding the owner to those inventories, so I updated the naming to be more accurate.

PR #532, opened 2025-04-03
chore: improve docker setup

0xxlegolas:
- improve docker build
- script to create test data

### DIFF
--- a/end-to-end-tests-v2/script/BulkCreateTestData.s.sol
+++ b/end-to-end-tests-v2/script/BulkCreateTestData.s.sol
@@ -23,35 +23,48 @@ import { EntityRecordParams, EntityMetadataParams } from "@eveworld/world-v2/src
 import { CreateInventoryItemParams } from "@eveworld/world-v2/src/namespaces/evefrontier/systems/inventory/types.sol";
 
 contract BulkCreateTestData is Script {
+  // Global variables for item IDs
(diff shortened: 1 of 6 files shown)


end-to-end-tests-v2/script/BulkCreateTestData.s.sol
```diff
@@ -0,0 +1,416 @@
...
+        y: 1001 + i,
+        z: 1001 + i
+      });
+
+      LocationData memory destinationGateLocation = LocationData({
+        solarSystemId: 1,
+        x: 1002 + i,
+        y: 1002 + i,
+        z: 1002 + i
+      });
```
ccp-bofai, line 251: Aren't these a bit close to each other? I am just wondering if we ever add logic for some sort of min/max distance that this might be an issue (I am questioning if the data is realistic, or if it looks like something that should be a failure case on grounds that we are not currently checking).

ccp-bofai (approved):
Looks good to me :) thanks for adding the `.dockerignore` file

end-to-end-tests-v2/script/BulkCreateTestData.s.sol
0xxlegolas: These are purely for local testing and unrealistic values. However we are not checking if they are close on-chain as well.

ccp-raudur (approved):

PR #533, opened 2025-04-07
refactor: update SmartGate system to include character ID in canJump

0xxlegolas:
- updated canJump function to include characterId 
- updated abis

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,5 +1,4 @@
 # Version control
-.git
 .gitignore
 .gitmodules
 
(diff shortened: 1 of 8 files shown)


.dockerignore
```diff
@@ -1,5 +1,4 @@
 # Version control
-.git
```
vayan, line 2: you sure?

vayan (approved):

.dockerignore
0xxlegolas: yes it fixes a build issue

ccp-raudur (approved):

PR #539, opened 2025-04-09
add test script for end-to-end-tests-v2, cleanup end-to-end-tests-v2,…

fc-tezza:
Change Log -

- Make end-to-end-tests-v2 testable by adding test script and workflow
- Adjust script calling pattern for “validated” calls (via the world.callFrom pattern) 
- Make cross-script variables consistent (use the already existing variables and state from previously runs scripts consistently) 
- Enforce singleton ownership assignment `revert` for InventoryOwnershipSystem - singleton items cannot already be assigned to another inventory of owned directly if they are going to be assigned.  Also adjust similar for OwnershipSystem - add Inventory assignment checking
- Remove direct system lib `view` calls from `InventoryInteract` and `EphemeralInteract`
- Update `EphemeralInteractSystem.transferFromEphemeral` to allow the ephemeral owner to interact by default
- Update world-v2 test for the above access changes
- Adjust depositFuel permissions


end-to-end-tests-v2/package.json
```diff
@@ -7,31 +7,33 @@
...
...--fork-url $RPC_URL --private-key $PRIVATE_KEY --broadcast --sig \"run(address)\" $WORLD_ADDRESS -vvv > test.log && forge script script/WithdrawFromEphemeral.s.sol:WithdrawFromEphemeral --fork-url $RPC_URL --private-key $PRIVATE_KEY --broadcast --sig \"run(address)\" $WORLD_ADDRESS -vvv > test.log && forge script script/TransferItems.s.sol:TransferItems --fork-url $RPC_URL --private-key $PRIVATE_KEY --broadcast --sig \"run(address)\" $WORLD_ADDRESS -vvv > test.log && forge script script/AnchorSmartTurret.s.sol:AnchorSmartTurret --fork-url $RPC_URL --private-key $PRIVATE_KEY --broadcast --sig \"run(address)\" $WORLD_ADDRESS -vvv> test.log && forge script script/ConfigureSmartTurret.s.sol:ConfigureSmartTurret --fork-url $RPC_URL --private-key $PRIVATE_KEY --broadcast --sig \"run(address)\" $WORLD_ADDRESS -vvv > test.log && forge script script/AnchorSmartGate.s.sol:AnchorSmartGate --fork-url $RPC_URL --private-key $PRIVATE_KEY --broadcast --sig \"run(address)\" $WORLD_ADDRESS -vvv > test.log && forge script script/ConfigureSmartGate.s.sol:ConfigureSmartGate --fork-url $RPC_URL --private-key $PRIVATE_KEY --broadcast --sig \"run(address)\" $WORLD_ADDRESS -vvv > test.log && rm test.log",
```
0xxlegolas: Since this runs with run-scripts.sh, could we move these commands into that script ? That way, we can group the Forge-related logic into a reusable function and keep short

end-to-end-tests-v2/package.json
```diff
@@ -7,31 +7,33 @@
...
+    "anchorSmartGate:local": "source ./.env.local && WORLD_ADDRESS=$WORLD_ADDRESS && RPC_URL=$RPC_URL && PRIVATE_KEY=$PRIVATE_KEY forge script script/AnchorSmartGate.s.sol:AnchorSmartGate --fork-url $RPC_URL --private-key $PRIVATE_KEY --broadcast --sig \"run(address)\" $WORLD_ADDRESS -vvv",
+    "configureSmartGate:local": "source ./.env.local && WORLD_ADDRESS=$WORLD_ADDRESS && RPC_URL=$RPC_URL && PRIVATE_KEY=$PRIVATE_KEY forge script script/ConfigureSmartGate.s.sol:ConfigureSmartGate --fork-url $RPC_URL --private-key $PRIVATE_KEY --broadcast --sig \"run(address)\" $WORLD_ADDRESS -vvv",
     "runAll:local": "pnpm run createSmartCharacter:local && pnpm run createMintERC20:local && pnpm run anchorSSU:local && pnpm run depositFuel:local && pnpm run bringOnline:local && pnpm run depositToInventory:local && pnpm run depositToEphemeral:local && pnpm run withdrawFromInv:local && pnpm run withdrawFromEphemeral:local && pnpm run transferItems:local && pnpm run anchorSmartTurret:local && pnpm run configureSmartTurret:local && pnpm run anchorSmartGate:local && pnpm run configureSmartGate:local",
```
0xxlegolas, line 28: Can we remove this since runAll already covers the same test?

end-to-end-tests-v2/package.json
```diff
@@ -7,31 +7,33 @@
...
     "simulateMetaTxn:local": "source ./.env.local && PRIVATE_KEY=$PRIVATE_KEY && FORWARDER_ADDRESS=$FORWARDER_ADDRESS && WORLD_ADDRESS=$WORLD_ADDRESS && forge script script/SimulateMetaTxn.s.sol:SimulateMetaTxn --fork-url $RPC_URL --private-key $PRIVATE_KEY --broadcast --sig \"run(address)\" $WORLD_ADDRESS -vvvv",
     "transferERC20:local": "source ./.env.local && PRIVATE_KEY=$PRIVATE_KEY && WORLD_ADDRESS=$WORLD_ADDRESS && forge script script/TransferERC20.s.sol:TransferERC20 --fork-url $RPC_URL --private-key $PRIVATE_KEY --broadcast --sig \"run(address)\" $WORLD_ADDRESS -vvvv",
     "lint": "pnpm run prettier && pnpm run solhint",
     "prettier": "prettier '**/*.sol' --plugin=prettier-plugin-solidity --check",
     "fixFormat": "prettier '**/*.sol' --plugin=prettier-plugin-solidity --write",
-    "solhint": "solhint --config ./.solhint.json 'src/**/*.sol'"
+    "solhint": "solhint --config ./.solhint.json 'src/**/*.sol'",
+    "test": "chmod +x run-scripts.sh && ./run-scripts.sh"
```
0xxlegolas, line 35: Can we use this to run all tests via a run-test.sh script? The script could:
1.  Deploy and configure the world
2.  Run all the Forge scripts in sequence
3.  Add some helpful logs
Then the test command can just call this script, which keeps package.json cleaner

end-to-end-tests-v2/run-scripts.sh
```diff
@@ -0,0 +1,62 @@
...
+export GATE_TYPE_ID=5
+export GATE_VOLUME=10000
+
+# Run deployments and tests
+
+echo "Running world-v2 upgrade..."
+pnpm nx run @eveworld/world-v2:upgrade || { echo "Upgrade failed"; kill $ANVIL_PID; exit 1; }
+
+echo "Running post-deploy..."
+pnpm nx run @eveworld/world-v2:post-deploy || { echo "Post-deploy failed"; kill $ANVIL_PID; exit 1; }
```
0xxlegolas: curious to know if there is any reason to split into `world-v2:upgrade` and `post-deploy` instead of `world-v2:deploy` ?

end-to-end-tests-v2/script/AnchorSSU.s.sol
```diff
@@ -62,7 +61,15 @@ contract AnchorSSU is Script {
       locationData: locationParams
     });
 
-    smartStorageUnitSystem.createAndAnchorStorageUnit(deployableParams, storageCapacity, ephemeralCapacity);
+    // createAndAnchorStorageUnit is a validated call, validated calls must be made from the deployer account via delegation using world.callFrom
```
0xxlegolas, line 64: 🙏

0xxlegolas:
the sof-state.json file is duplicated. Should we move to a project root or smart-object-framework folder so that end-end-test and world can use it ?

end-to-end-tests-v2/run-scripts.sh
```diff
@@ -0,0 +1,62 @@
...
+
+echo "Running config..."
+pnpm nx run @eveworld/world-v2:config || { echo "Config failed"; kill $ANVIL_PID; exit 1; }
+
+echo "Running scripts..."
+pnpm nx run end-to-end-tests-v2:runAll:test || { echo "Scripts failed"; kill $ANVIL_PID; exit 1; }
+
+# Kill anvil process
+echo "Tests completed. Shutting down Anvil."
+kill $ANVIL_PID
```
0xxlegolas: [run-testssh.md](https://github.com/user-attachments/files/19665758/run-testssh.md)
could not upload a .sh file. Here is a ref

end-to-end-tests-v2/script/TransferItems.s.sol
```diff
@@ -221,9 +191,26 @@ contract TransferItems is Script {
...
+      abi.encodeCall(FuelSystem.depositFuel, (ssuSmartObjectId, 1000))
+    );
+
+    world.callFrom(
+      invOwner,
+      deployableSystem.toResourceId(),
+      abi.encodeCall(DeployableSystem.bringOnline, (ssuSmartObjectId))
+    );
   }
 }
```
0xxlegolas, line 216: 🙏 thanks for the heavy lift. So the issue was I did not use the callFrom before ?

end-to-end-tests-v2/package.json
fc-tezza: I have removed the `runAll:test `now and put this logic into the `.sh `script itself

end-to-end-tests-v2/package.json
fc-tezza: Good suggestion, I have made this change now.

end-to-end-tests-v2/run-scripts.sh
fc-tezza: Thanks, I've used this.

end-to-end-tests-v2/script/TransferItems.s.sol
fc-tezza: I set these to `callFrom` since they are "validated" calls (technically `bringOnline` is NOT validated, but still accepts `_callMsgSender(1)` as an owner, so it fits within this pattern too).

I think the issue was this commented out deployer `vm.startPrank` in the previous version - https://github.com/projectawakening/world-chain-contracts/blob/8e47418e9a8138eb03acfce7945a996e2115d01e/end-to-end-tests-v2/script/TransferItems.s.sol#L69

.gitignore
```diff
@@ -22,3 +22,6 @@
 # Ignore all env files
 **/.env
 .pnpm-store/
+
+# Ignore logs
+**/logs/**
```
0xxlegolas, line 27: good catch, can we also include this in ..dockerignore ?

0xxlegolas (approved):

PR #540, opened 2025-04-10
fix: replace unavailable Foundry image with Ubuntu base image

0xxlegolas:
fix: replace unavailable Foundry image with Ubuntu base image

- Previous approach used a pinned Foundry image (SHA: 8b843eb) that is no longer available
- Switched from Alpine-based environment to Ubuntu for better compatibility
- Maintains the same functionality while ensuring build reliability

### DIFF
--- a/build/package/deploy-v2.sh
+++ b/build/package/deploy-v2.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 set -eou pipefail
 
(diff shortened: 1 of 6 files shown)


ccp-bofai (approved):
Looks good to me

vayan (approved):

tinkrtailor (approved):

PR #564, opened 2025-05-05
Feature/network node & energy decoupling

0xxlegolas:
**Changes**

- Decoupled fuel logic from all smart structures.
- Removed globalOnline and globalOffline—these aren't used by Admin. Use world-level pause/resume when the game goes offline. 
- Implemented new fuel logic that burns fuel based on fuel efficiency.
- Added fuel tracking: consumption status, burn state, and time remaining for the current fuel amount (requires a cron job at regular intervals).
- Introduced Network Node system: structures can optionally connect via a networkNodeId to draw energy from the Network Node.
- Added check for energy requirement when bringing a structure online—only possible if connected to a Network Node with sufficient energy.

**summary :**
The Network Node Energy System is now a central part of the base-building mechanic. Network Nodes generate energy from fuel and power all nearby structures within a player’s base.

**user flow :** 
- Player deploys Network Node structure
- Player anchors Network Node at a location
- Player deposits fuel into Network Node
- When the Player burns the fuel, it consumes 1 unit of fuel, yeilds 'x' energy and last until the configured fuel consumption rate. eg: 1 hour
- Player brings Network Node online. It reduces the energy required from the total energry
- Player deploys additional structures (SSU, ST, Industry, etc.)
- Player anchors structures near Network Node
- Player brings structures online (if sufficient energy available from the total energy from the network node)
- If energy becomes insufficient, the structure cannot be brought online 
- If Network Node goes offline (no fuel), all structures go offline

**other things to consider :** 
- fuelEfficiency: should be between 10 and 100.
- fuelBurnRate: must be greater than 1 minute.


vayan:
> network node energy decoupling

cool PR name btw

spacetime-navigator (dismissed):

fc-tezza (approved):
I think there are improvements to be made in future iterations, but for a first cut LGTM!

0xxlegolas:
@ccp-tezza Do you a suggestions/comments for the next iteration ? I would love to consider that in the next PR ?

PR #565, opened 2025-05-06
chore: batch call system tag script

fc-tezza:
Added a batch call system tag script.

This script adds the batch call tag to all base smart assembly classes (specifically for the Nova tenant). 
 
This is a patch for any existing 0.1.2 world deployments (or older) allowing us to use the batch call system in those environments. 

NOTE: this logic is applied by default to any fresh world deployment of 0.1.3 or later, and hence the script is not needed in those environments.

To run the script navigate to the world-chain-contracts root folder and run
`pnpm --filter @eveworld/smart-object-framework-v2 update:batch-call-tag`

This script assumes you have the world deployer key accessible as an ENV variable


0xxlegolas (commented):
LGTM !

0xxlegolas (approved):

PR #566, opened 2025-05-12
Refactor/network node operation

0xxlegolas:
- merging burn and bringOnline in single function for NetworkNode
- removed bunch of params from createAndAnchorNetworkNode 
- added more params in depositFuel
- changed fuelTypeId  as smartObjectId to be consistent with other ids. 
- fixes naming convention from structures to assemblies


mud-contracts/world-v2/src/codegen/world/INetworkNodeSystem.sol
```diff
@@ -27,11 +27,11 @@ interface INetworkNodeSystem {
...
   ) external;
 
-  function evefrontier__connectStructure(uint256 networkNodeId, uint256 structureId) external;
+  function evefrontier__connectStructure(uint256 networkNodeId, uint256 assemblyId) external;
 
-  function evefrontier__onStructureOnline(uint256 networkNodeId, uint256 structureId) external;
+  function evefrontier__onStructureOnline(uint256 networkNodeId, uint256 assemblyId) external;
 
-  function evefrontier__onStructureOffline(uint256 networkNodeId, uint256 structureId) external;
+  function evefrontier__onStructureOffline(uint256 networkNodeId, uint256 assemblyId) external;
```
vayan: do we also want to rename those? 

`evefrontier__connectAssembly`
`evefrontier__onAssemblyOnline`
`evefrontier__onAssemblyOffline`

0xxlegolas: yes on it changing one by one haha

fc-tezza (approved):

PR #567, opened 2025-05-12
refactor: export sof-v2 scripts as pure functions

dk1a:
I separated the body of what scripts do (calling systems, setting tables) into separate functions, which scripts then call;
scripts remain effectively unchanged, but their standalone functions can now also be imported

For my setup it's more convenient to import a function/library, than to run `forge script` or `--load-state`.
Specifically I run all the setup scripts for local tests within `setUp`, since that's much faster


0xxlegolas:
yes totally agreed , importing a function/library is the best way to go. 

when you say `setUp` is it part of the postDeploy scripts ?

dk1a:
> when you say `setUp` is it part of the postDeploy scripts ?

it's both depending on the situation

tests run here:
https://github.com/dk1a/wanderer-cycle/blob/e05d35b19d3af0a7aa3d03f707b8459078f2f6f0/packages/contracts/test/BaseTest.t.sol#L25

but actual PostDeploy also exists calling the same function
https://github.com/dk1a/wanderer-cycle/blob/e05d35b19d3af0a7aa3d03f707b8459078f2f6f0/packages/contracts/script/PostDeploy.s.sol#L97

but PostDeploy is only called for full deployments, not tests, because it takes several minutes (mud doesn't have a flag for this, so I work around that by using different script directories via `FORGE_SCRIPT`)

dk1a:
Also on second thought `address worldAddress` argument for the functions is rather unnecessary, even if you do need it later then `_world()` works; should I remove it?

fc-tezza (approved):
This makes a lot of sense. This way these scripts can be easily useable whichever approach you are using for test/deploy!

PR #568, opened 2025-05-13
add configure scripts

0xxlegolas:


.env
```diff
@@ -32,8 +32,13 @@ GATE_VOLUME=10000
...
-FUEL_TYPE_ID=7
-
+# Energy constant by assembly
+ASSEMBLY_TYPE_ID=[88092,84556,84955,88086,87160,88093,88094,88068,87162,87161,88067,87120,87119,88063,88064,88070,88071,88069,77033,88098,88099,88100,88101,77917,87566,88083,88082,88084,88095,88096,88097]
+ENERGY_CONSTANT=[0,10,100,50,0,0,0,20,0,0,20,30,10,10,20,20,30,10,0,0,0,0,100,0,0,0,0,0,0]
+
+# Fuel efficiency by fuel type
+FUEL_TYPE_ID=["77818","78437","78515","78516","84868"]
+FUEL_EFFICIENCY=["20","90","80","40","40"]
+FUEL_VOLUME=["28","28","28","28","28"]
```
vayan: a bit fragile to rely on matching array sizes, we'll probably want something more advanced if we continue using that! fine for now 😄

vayan (approved):

.env
0xxlegolas: adding some checks to it to make sure array size is same

fc-tezza (approved):

PR #569, opened 2025-05-13
feat: add an installation library for SOF-v2

dk1a:
Depends on #567 (InstallSOFLib uses those pure functions)

The SOF installation can be run multiple times, and upgrades systems on reruns, kinda like `mud deploy`.
This means that adding it to PostDeploy in parallel to other scripts shouldn't be an issue, and you could also simplify some deploy scripts later, but I haven't tested it everywhere in your codebase.
I added an extra test pipeline in `world-v2`, otherwise I'm not sure how far you want to integrate this approach.

A few minor issues I can see:
- stale addresses in CallAccess, I noted that in a TODO
- it will upgrade all its systems aggressively even if nothing changed, perhaps this could be improved via using a module and/or deterministic deployment
- it would be nice for all of sof stuff to have a separate namespace unrelated to evefrontier

I've considered making it a module, but that leaves more complicated code due to delegation, and probably won't work due to `callCount == 1` in `AccessConfigSystem` (that could be refactored I suppose). I also didn't see much benefit of the module.
I based my code style on https://github.com/latticexyz/mud/blob/main/packages/world-module-metadata/src/MetadataModule.sol (it's one of the newer modules), and there you can also see the delegation approach (`callFrom`)


dk1a:
tentatively closing in favor of #574

PR #570, opened 2025-05-14
fix: add missing access control to tag methods

dk1a:
An access issue I noticed while looking at the systems

I think these, as well as `deleteObjects`/`deleteClasses`, don't have any access tests. The singular funcs already look well tested, but the plural wrappers could use a sanity check too


fc-tezza (approved):
Good catch!

you, are correct these multi-Tag functions are supposed to call their protected single tag counterparts (which have access control implemented) rather than calling the internal `_setTag`/`_removeTag` logic function.

The `createObjects` and `deleteObjects` function in the EntitySystem already call their access controlled counterparts and so are safe in this regard.

dk1a:
> you, are correct these multi-Tag functions are supposed to call their protected single tag counterparts (which have access control implemented) rather than calling the internal `_setTag`/`_removeTag` logic function.

Gotcha, though I think I phrased my comment poorly, it was mostly meant as a suggestion to add some small test to `SOFAccessSystemTest` that `setTags` etc have access control

PR #572, opened 2025-05-15
fix: less strict AccessConfigSystem namespace ownership check

dk1a:
After some experimentation with #569 it had more problems than I expected, so I'd rather do a module.
The module should ideally use delegation (juggling namespace ownership is much more dangerous), so AccessConfigSystem needs to support that

I don't think this introduces any risks, this ownership check is what mud's own access control system uses too


fc-tezza (approved):
Originally we restricted this to direct calling only (`callCount == 1`) because that was the only way were were calling it in our current workflows and we want to follow the Principle of Least Privilege when it come to access control logic.

However, I agree, this can help with being able to use the delegation pattern and make the System more useful generally.

You are also correct that it does not really change the security profile of the contract to anything that is not already a standard pattern in the MUD core logic.

In the context of this codebase, since we are always using the `msgSender` of the current call in the execution callstack, then the new logic is equivalent to the old logic for `callCount == 1` but still requires the caller to be the `NamespaceOwner` at any call depth and so will remain safe.

PR #573, opened 2025-05-15
chore: update mud for sof-v2 and world-v2

dk1a:
Related to #569 and #572 
It's not necessary, but module code is cleaner with newer mud versions; unless you have reasons to keep 2.2.15


ccp-tezza:
Upgrading to the latest version of MUD is a good idea, but before we do that we need to make sure of a few things to ensure the newest mainline version supports current workflows:
- does the update introduce other changes, and if it does do need to adjust the codebase to accommodate 
- does the latest mainline support System Libs auto-gen (we use this extensively, and need to make sure the new mainline supports it)
- confirm if any game infra integration points are going to be affected by the update
- if we update the MUD version, it should be uniformly updated across the whole repository.. this includes the following packages:
    - mud-contracts/core-v2
    - mud-contracts/smart-object-framework-v2
    - mud-contracts/world-v2
    - end-to-end-tests-v2
    - standard-contract-v2

dk1a:
I updated the mentioned packages.
`standard-contract-v2` doesn't use mud and is unaffected.
Tests passed fine, except e2e, but it doesn't work for me on `develop` either, I think `.env.local` is outdated; tests there also don't terminate `anvil` on failure, which caused some further confusing errors

The only breaking change I noticed is in `SmartObjectFramework`, which is why I added a workaround to it.
Otherwise it's a lot of explorer/entrykit changes (you don't use it), and some module and system lib improvements.
Of course I don't know much at all about your infra and you'll want someone from the team to check that.

System libs are supported in all the latest mud versions, the updates only improved them, and I imagine will keep doing so, it's still kind of an experimental feature.
In fact part of the reason for the update is that `@latticexyz/world` now exports some core system libs so you can use them instead of e.g. `world.registerNamespace` (useful for module)

PR #578, opened 2025-05-19
refactor: add separate _enforceScope sof method

dk1a:
Using `enforceCallCount` for naming style; or would `_scopeGuard` be better?

Useful for when I have a method that operates on 2 separate entities, and I have to enforce scope on both.
I think just using a function makes more sense than adding more modifier overrides (especially since modifier overrides don't work and they'll need different names)


fc-tezza (approved):
Thanks this is a good suggestion!

In fact were already considering downgrading `scope()`  from a modifier into an internal call (at least in the world-v2 implementation). 

So far in our usage of the SOF, it turns out that the scoping model adds a lot of configuration complexity (tagging scoped systems to a class or object ) and adds an `extra` layer of call access scoping checks on-top of the already existing `access()` checks. Since these patterns are so similar it might better to just combine them at some point (handling everything as part of a streamlined `access()` flow. 

This makes a lot of sense considering that `scope()` is just a special case of the wider access control pattern.

Originally, we also exposed `scope()` separately as a modifier as we thought it would help with code readability, but it seems that through some streamlining of the `access()` flow, the call restriction outcomes of scoping may be able to be achieved in a simpler fashion.

This possible "streamlining" change is still an on-going conversation. Would be happy to get your thoughts on it at some point.

PR #579, opened 2025-05-19
refactor: improve inheritance for sof-v2 systems

dk1a:
I want `registerClass` and `instantiate` to create and return their `classId`/`objectId`, rather than accept it as argument. This would be trivial if `EntitySystem` had virtual methods and an internal `_registerClass`/`_instantiate`.
So instead of duplicating all the code, I want to be able to do this: https://github.com/dk1a/wanderer-cycle/blob/sof/packages/contracts/src/namespaces/evefrontier/EntitySystemTODOInherit.sol

Other systems also lose nothing from making inheritance easier.

I'm leaving `SmartObjectFramework`/`SOFAccessSystem` out of this for now, since I'm not sure about it, nor do I need to override most of their methods.


fc-tezza (approved):
Yes, the intention is for the SOF to be fully customizable by users of it. Customization via override is a great addition here.

ccp-tezza:
@dk1a there are some linter errors blocking a merge of this.

Could you run `pnpm nx run-many -t fixFormat` from the root directory in your branch and push those changes to help unblock?

dk1a:
Updated

PR #584, opened 2025-05-26
feat: add MockForwarder for system-to-system tests

dk1a:
I didn't want to manually create mocks for every system test that I'm refactoring to SOF. I don't really like having to rely on world selectors, but this forwarding wouldn't work for system libs atm

I refactored `SOFAccessSystem.t.sol` just as a showcase, the main benefit would be in losing all the mocks that projects using SOF have to maintain, like the ones in `world-v2`.
If you like this example I can clean it up more and add some tests for the mock forwarder itself


ccp-tezza:
Will take a look, thanks!

PR #590, opened 2025-05-29
fix: update fuel consumption logic to handle zero fuel and elapsed time

0xxlegolas:
Add support for bringing systems online without fuel when previous cycle time remains


ccp-raudur (approved):
Perfect

PR #595, opened 2025-05-30
fix deployer image

0xxlegolas:
- Pinned a specific Foundry version to resolve issues with the deployer image.
- Added environment variable validation to ensure required values are set. The script now fails gracefully with clear messages before execution if validations aren't met.


build/package/deploy-v2.sh
```diff
@@ -14,6 +14,42 @@ trap 'cp $LOG_FILE "logs/$(date +%Y%m%d_%H%M%S)-deploy-in-docker-v2.log"' EXIT
 # Parse command line arguments
 parse_arguments "$@"
 
+# Function to validate and log environment variables
+validate_and_log_env() {
```
vayan, line 18: so nice, thanks :pray:

ccp-raudur (approved):
🙏

build/package/deploy-v2.sh
0xxlegolas: @vayan  with the latest changes, the .env also requires changes. It affects the downstream image deployment. 
It does not break but the `TYPE_IDS` should be updated else tests might fail.

PR #621, opened 2025-06-09
feat: improve fuel handling

0xxlegolas:
Ignores the last unit of fuel when a user deposits a different fuel type than the currently burning one.


ccp-raudur (approved):
Great!

PR #628, opened 2025-06-27
**fix**: Add  No Access permission to transferToInventory

fc-tezza:
Correctly add the `NoAccess` permissions to the `InventoryInteract.transferToInventory` call. 

This functionality is here to lay the groundwork for future location bounded uses cases between separate Smart Assembly inventories.
 
It now correctly has the `No Access` permission level until we adjust it later with enforcement of location bounded digital physics rules


0xxlegolas (approved):
LGTM

PR #757, opened 2025-12-18
bugfix : prevent energy release when destroying offline assemblies

0xxlegolas:
**Bug :** 
Energy was released from network nodes when destroying assemblies that were not consuming energy (i.e., in ANCHORED state). This occurred during destroyDeployable and unanchor operations.

Fixed: 
Added a state check to release energy only when the assembly was previously ONLINE. Assemblies in ANCHORED state no longer trigger energy release during destruction.


ccp-raudur (approved):
LGTM

Issue #756, opened 2025-12-18
Network Node anchored, offline Assembles still release reserved energy from overall Network Node


ProtoDroidBot:
When a Deployable requiring energy to function is unanchored/destroyed when in the Anchored (non-Online) state, it still releases the energy from the network node:

Traced the issue back to two functions in the DeployableSystem.sol contract file, there's no explicit check on the deployable if its actually online or not after the initial check, before releasing the energy.

https://github.com/projectawakening/world-chain-contracts/blob/develop/mud-contracts/world-v2/src/namespaces/evefrontier/systems/deployable/DeployableSystem.sol#L298 (Unanchor function)

https://github.com/projectawakening/world-chain-contracts/blob/develop/mud-contracts/world-v2/src/namespaces/evefrontier/systems/deployable/DeployableSystem.sol#L165 (Destroy function)

ProtoDroidBot:
Whelp, following commit resolves the issue, thanks! https://github.com/projectawakening/world-chain-contracts/commit/1a5c18395224b6ebb81cab6bfd3b0556e62b67f9

0xxlegolas:
Thanks for finding the exploit

ProtoDroidBot:
Now I am unsure how this exact issue works, as if ReleaseNetworkEnergy is called by the above two functions (Unanchor and Destroy), then the check for if the deployable is generating power is happening in that function, not the DeployableSystem Function: https://github.com/projectawakening/world-chain-contracts/blob/develop/mud-contracts/world-v2/src/namespaces/evefrontier/systems/network/NetworkNodeSystem.sol#L233


```
   if (releasedEnergy > 0) {
      uint256 currentReserved = NetworkNode.getTotalReservedEnergy(networkNodeId);
      uint256 newReserved = currentReserved > releasedEnergy ? currentReserved - releasedEnergy : 0;
```

@0xxlegolas would it make sense to have a sanity check on this before I completely close out this thread?

0xxlegolas:
@ProtoDroidBot  Are you saying to add a extra check 
```
currentReserved >= releasedEnergy
```

ProtoDroidBot:
@0xxlegolas for the most part, I am wary about adding new things if existing systems are taking care of this already. These might make sense, but its CCP's call

PR #759, opened 2025-12-19
fix: allow assembly disconnection regardless of state

0xxlegolas:
Issue: #756 introduced a bug where only online assemblies could be disconnected, preventing disconnection of anchored assemblies.

**Fix:** Decoupled energy release from disconnection. disconnectAssembly now runs regardless of state, while releaseAssemblyEnergy only runs when the assembly is online.

Changes:
- Energy release remains conditional on ONLINE state only


ccp-raudur (approved):
Makes sense

